PR #5, opened 2016-12-28
Zero cost futures via associated type

### LINKED ISSUE #3: Get rid of namespace traits?
Doesn't bring in much value to have them, and you need to import them :/

rphmeier:
Also closes #3 

Transports will now expose an `Out` future type which can be composed with a deserializing `CallResult` future (as a stopgap until `impl Trait`)

Hurts ergonomics slightly, but a future PR will allow manual erasure of the transport type.

### DIFF
--- a/examples/http.rs
+++ b/examples/http.rs
@@ -5,7 +5,6 @@ extern crate web3;
 
 use std::thread;
 use futures::Future;
-use web3::api::Eth;
 
 fn main() {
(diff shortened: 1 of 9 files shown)


tomusdrw:
:tada: :tada: :tada: :tada: :tada: :tada: :tada: :tada: :tada: :tada: :tada: :tada: :tada: :tada: :tada: :tada: :tada: :tada: 

:raised_hands:

PR #9, opened 2016-12-30
re-export futures crate from lib

### LINKED ISSUE #8: Re-export Futures as submodule
As `futures-rs` is part of the `rust-web3` public API it should be re-exported.

rphmeier:
closes #8

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -2,7 +2,6 @@
 
 #![warn(missing_docs)]
 
-extern crate futures;
 extern crate jsonrpc_core as rpc;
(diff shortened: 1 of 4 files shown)


src/lib.rs
```diff
@@ -23,6 +22,13 @@ pub mod transports;
 
 pub use api::{Web3Main as Web3, ErasedWeb3};
 
+/// Re-export of types from the `futures` crate.
+pub mod futures {
+  extern crate futures;
```
tomusdrw: Why not just `pub extern crate futures`?

PR #10, opened 2017-01-02
IPC transport

tomusdrw:
`tokio_uds` based IPC client


src/transports/http.rs
```diff
@@ -65,22 +65,22 @@ impl Transport for Http {
...
-	url: String,
-	client: Arc<reqwest::Client>,
+  url: String,
+  client: Arc<reqwest::Client>,
   request: String,
 }
 
 impl Future for FetchTask {
-	type Item = rpc::Value;
-	type Error = RpcError;
```
rphmeier: haha, sorry about messing up the indentation here :)

(do you think we could switch to 4spaces at some point? 2 is a bit tight)

tomusdrw: Absolutely https://github.com/tomusdrw/rust-web3/issues/12

src/transports/ipc.rs
```diff
@@ -0,0 +1,376 @@
...
+use self::tokio_core::io::{ReadHalf, WriteHalf, Io};
+use self::tokio_uds::UnixStream;
+
+use futures::{self, Sink, Stream, Future, BoxFuture};
+use futures::sync::{oneshot, mpsc};
+use helpers;
+use rpc::{self, Value as RpcValue};
+use {Transport, Error as RpcError};
+
+macro_rules! try_nb {
```
rphmeier: isn't there already a macro for this in the mio ecosystem somewhere?

src/transports/ipc.rs
```diff
@@ -0,0 +1,376 @@
...
+      return Err(())
+    },
+  })
+}
+
+/// Error returned while initializing IPC transport
+#[derive(Debug)]
+pub enum IpcError {
+  /// Wrapped IO error
+  Io(io::Error),
```
rphmeier: might as well just return an `io::Error` from the initialization if it's the only variant. this will make integration easier.

src/transports/ipc.rs
```diff
@@ -0,0 +1,376 @@
...
+    Ok(Ipc {
+      id: atomic::AtomicUsize::new(1),
+      write_sender: write_sender,
+      read_sender: read_sender,
+    })
+  }
+
+  /// Create new IPC transport with separate event loop.
+  /// NOTE: Dropping event loop handle will stop the transport layer!
+  pub fn with_event_loop<P>(path: P) -> Result<(EventLoopHandle, Self), IpcError> where
```
rphmeier: the `with_` naming convention implies that the thing will be provided to the function (like `with_capacity`)

So to me it would make more sense for you to switch the names of `new` and `with_event_loop`

src/transports/ipc.rs
```diff
@@ -0,0 +1,376 @@
...
+        ipc,
+      ))
+  }
+}
+
+impl Transport for Ipc {
+  type Out = BoxFuture<RpcValue, RpcError>;
+
+  fn execute(&self, method: &str, params: Vec<String>) -> Self::Out {
+    let id = self.id.fetch_add(1, atomic::Ordering::Relaxed);
```
rphmeier: wouldn't it be safer to use an `AcqRel` or better yet `SeqCst` here? I think `fetch_add` is always interlocked on x86 but not sure about other platforms.

src/transports/ipc.rs
```diff
@@ -0,0 +1,376 @@
...
+    buffer: Vec<u8>,
+    current_pos: usize
+  },
+}
+
+/// Writing part of the IPC transport
+/// Awaits new requests using `mpsc::Receiver` and writes them to the socket.
+struct WriteStream {
+  write: WriteHalf<UnixStream>,
+  incoming: mpsc::Receiver<String>,
```
rphmeier: why receive utf-8 strings? Seems that a byte-vec is more accurate.

src/transports/ipc.rs
```diff
@@ -0,0 +1,376 @@
...
+      warn!("Got unsupported response (id: {:?})", id);
+    }
+  }
+
+  fn extract_response(buf: &[u8], min: usize) -> Option<(rpc::Output, usize)> {
+    for pos in (min..buf.len()).rev() {
+      // Look for end character
+      if buf[pos] == b']' || buf[pos] == b'}' {
+        // Try to deserialize
+        match helpers::to_response_from_slice(&buf[0..pos + 1]) {
```
rphmeier: if `pos == buf.len() - 1` here then this will be OOB, won't it?

src/transports/ipc.rs
```diff
@@ -0,0 +1,376 @@
...
+impl Future for ReadStream {
+  type Item = ();
+  type Error = ();
+
+  fn poll(&mut self) -> futures::Poll<Self::Item, Self::Error> {
+    // Check new incoming response channels
+    let response = self.incoming.poll();
+    if let Ok(futures::Async::Ready(Some((id, response)))) = response {
+      self.pending.insert(id, response);
+    }
```
rphmeier: shouldn't it drain all incoming IDs here?
A shared `Arc<Mutex<BTreeMap>>` might have lower overhead here

src/transports/ipc.rs
```diff
@@ -0,0 +1,376 @@
...
+        return Ok(futures::Async::NotReady);
+      }
+
+      let min = self.current_pos;
+      self.current_pos += read;
+      if let Some((output, len)) = Self::extract_response(&self.buffer[0..self.current_pos], min) {
+        // copy rest of buffer to the beginning
+        for i in len..self.current_pos {
+          self.buffer.swap(i, i - len);
+        }
```
rphmeier: `ptr::copy` (_not_ non-overlapping) could definitely make this into as few copies as possible. But odds are it will all fit into the L1 cache anyway and won't be a bottleneck.

rphmeier: And LLVM probably can optimize it as well

rphmeier:
A couple of grumbles but mostly fine. 
I wonder if in the future it would be possible to use a concrete output future type for the transport but that's definitely an optimization for later.

src/transports/ipc.rs
tomusdrw: It does `return Err(e.into())` I'm just logging warnings instead.

src/transports/ipc.rs
tomusdrw: Slice `&buf[0..buf.len()]` should work fine, no?

src/transports/ipc.rs
tomusdrw: Right, there might be some responses pending in the channel. I will change it back to `Mutex`.

src/transports/ipc.rs
rphmeier: right, this should be fine then

tomusdrw:
Grumbles addressed :)

PR #18, opened 2017-01-10
Contract calls

tomusdrw:
Example:
```rust
let eth = web3.eth();
let abi = include_bytes!("./res/token.json");
let address = 1.into();
let token = Contract::from_json(eth, address, abi).unwrap()

token.query("name", (), None, None).wait().unwrap()
assert_eq!(result, "Hello World!".to_owned());
```


tomusdrw:
Hah, almost forgot about this PR :)

PR #20, opened 2017-01-22
More types: block, transactions, receipt, and log.

rphmeier:


src/api/eth.rs
```diff
@@ -366,14 +422,16 @@ mod tests {
     Eth:block:block_by_hash, BlockId::Hash(0x123.into()), true
     =>
     "eth_getBlockByHash", vec![r#""0x0000000000000000000000000000000000000000000000000000000000000123""#, r#"true"#];
-    Value::Null => ()
+    ::serde_json::from_str(EXAMPLE_BLOCK).unwrap()
+    => ::serde_json::from_str::<Block>(EXAMPLE_BLOCK).unwrap()
```
rphmeier: side note -- testing deserialization isn't really valuable for these tests. we're more interested in whether the right _call_ gets dispatched.

src/types/transaction.rs
```diff
@@ -0,0 +1,58 @@
...
+  /// Recipient (null when contract creation)
+  pub to: Option<H160>,
+  /// Transfered value
+  pub value: U256,
+  /// Gas Price
+  #[serde(rename="gasPrice")]
+  pub gas_price: U256,
+  /// Gas amount
+  pub gas: U256,
+  /// Inpur data
```
tomusdrw: typo

src/types/transaction.rs
```diff
@@ -0,0 +1,58 @@
...
+  pub hash: H256,
+  /// Nonce
+  pub nonce: U256,
+  /// Block hash. Null when pending.
+  #[serde(rename="blockHash")]
+  pub block_hash: Option<H256>,
+  /// Block number. Null when pending.
+  #[serde(rename="blockNumber")]
+  pub block_number: Option<U256>,
+  /// Transaction Index. Null when pending.
```
tomusdrw: `None when pending`?

src/types/block.rs
```diff
@@ -1,5 +1,104 @@
...
+  /// Timestamp
+  pub timestamp: U256,
+  /// Difficulty
+  pub difficulty: U256,
+  /// Total difficulty
+  #[serde(rename="totalDifficulty")]
+  pub total_difficulty: U256,
+  /// Seal fields
+  #[serde(rename="sealFields")]
+  pub seal_fields: Vec<Bytes>,
```
tomusdrw: Should it be `Option<Vec<Bytes>>` or will it default to empty `Vec` if the field is not there?

src/api/eth.rs
```diff
@@ -424,7 +482,8 @@ mod tests {
     Eth:transaction:tx_by_hash, TransactionId::Hash(0x123.into())
     =>
     "eth_getTransactionByHash", vec![r#""0x0000000000000000000000000000000000000000000000000000000000000123""#];
-    Value::Array(vec![]) => Some(())
+    ::serde_json::from_str(EXAMPLE_TX).unwrap()
```
tomusdrw: maybe `use serde_json` to avoid those `::`?

src/api/eth.rs
tomusdrw: Right, for simple structures it indeed might not make much sense, but still choosing right params to deserialize from is important and it's part of the logic, and some methods will do different things depending on the output.

Maybe instead of deserializing structs from strings we should just put `Default::default()`?

rphmeier: but then we'd also have to serialize the default struct to value -- not a huge win, but you might be right

src/types/block.rs
rphmeier: I think as things are right now it will fail to deserialize on the field missing. I suppose it could be optional, but then we'd also have to support non-generic `nonce` and `mixHash` fields.

rphmeier:
For some reason I can't respond directly to your comment which said this, but the issue with passing `Default::default()` is that we must also provide a corresponding `Value` which is equivalent. So either we deserialize twice or serialize once -- one isn't much better than the other. We could also just take `None` in each case...

src/types/block.rs
```diff
@@ -1,5 +1,104 @@
...
+    }
+  }
+}
+
+impl Deserialize for BlockTransactions {
+  fn deserialize<D>(deserializer: &mut D) -> Result<Self, D::Error>
+    where D: Deserializer
+  {
+    let res = Vec::<Transaction>::deserialize(deserializer).map(BlockTransactions::Full)
+      .or_else(|_| Vec::<H256>::deserialize(deserializer).map(BlockTransactions::Hashes));
```
rphmeier: so this line seems not to work as expected -- it only works if the type of transaction you're trying to get comes first. My guess is that the calls of the first attempt's `visit_*` leave the deserializer in an invalid state to deserialize the other type. We might just want to make `Block` generic over a transaction type and change the `eth` API to have two calls: 

`block(&self, BlockId) -> Block<H256>` and 

`block_with_txs(&self, BlockId) -> Block<Transaction>`

(uncle functions will return `Block<H256>`)

tomusdrw: I like splitting the API better.

Issue #13, opened 2017-01-03
Implement missing standard APIs


tomusdrw:
Have a look at the list in README

jellelicht:
Will shh also be supported eventually?

rphmeier:
@jellelicht While the whisper API is far from standard, work on implementing it is in-progress.

tomusdrw:
Sure, feel free to update the Readme or submit a PR with the code.
Also note that you can implement your own namespace and use generic `web3.api()` method: https://tomusdrw.github.io/rust-web3/web3/api/struct.Web3Main.html#method.api
to instantiate your custom methods.

folex:
I want to try myself on implementing `personal_sendTransaction`. It's not in the RPC spec though. Will you accept such PR?

folex:
Hi! I want to try implement `personal_signTransaction`, but it's not in the RPC spec. Will you accept such PR?

tomusdrw:
@folex yes, should be fine.

Issue #4, opened 2016-12-28
Better types module


rphmeier:
Probably can port over or publish part of `ethcore-rpc`

tomusdrw:
yup, already listed in todos, I think I would rather port them then trying to re-use.

rphmeier:
Partially addressed by #7 
Next up: Block and Transaction types.

rphmeier:
Done in #20

PR #25, opened 2017-02-23
Filters support and Work type

tomusdrw:
Related #13
Related #4


src/api/eth_filter.rs
```diff
@@ -1,10 +1,403 @@
...
+
+  fn uninstall_internal(&self) -> CallResult<bool, T::Out> {
+    let id = helpers::serialize(&self.id);
+    CallResult::new(self.transport.execute("eth_uninstallFilter", vec![id]))
+  }
+}
+
+impl<T: Transport, R> Drop for BaseFilter<T, R> {
+  fn drop(&mut self) {
+    let _ = self.uninstall_internal().wait();
```
rphmeier: does it need waiting? Should be enough just to dispatch, right?

tomusdrw: I think it needs to be polled couple of times (you cannot tell how many, cause it might depend on transport and io availability)

src/api/eth_filter.rs
```diff
@@ -1,10 +1,403 @@
...
+/// Result of installing new filter
+pub struct FilterResult<T, F, X> {
+  transport: T,
+  inner: F,
+  _output: PhantomData<X>,
+}
+
+impl<T, F, X> FilterResult<T, F, X> {
+  /// Create a new CallResult wrapping the inner future.
+  pub fn new(transport: T, inner: F) -> Self {
```
rphmeier: doc here is wrong

PR #28, opened 2017-05-06
Bump Deps, add Ord and From<&[u8]> to uint.

tomusdrw:


src/types/uint.rs
```diff
@@ -93,6 +115,16 @@ macro_rules! impl_uint {
       }
     }
 
+    impl<'a> From<&'a [u8]> for $name {
```
rphmeier: `T: AsRef<[u8]>` might be even better

src/types/uint.rs
```diff
@@ -93,6 +115,16 @@ macro_rules! impl_uint {
...
+      fn from(num: &'a [u8]) -> Self {
+        let mut arr = [0; $len];
+        for i in 0..cmp::min(num.len(), $len) {
+          arr[$len - 1 - i] = num[num.len() - 1 - i]
```
rphmeier: seems like this whole thing can be replaced with a `arr[..].copy_from_slice(&x[..len])` which will do a direct `memcpy` and not rely on LLVM optimizing the loop

rphmeier:
LGTM modulo nits; `AsRef` isn't strictly required but I think it makes sense here.

src/types/uint.rs
tomusdrw: Seems that it conflicts with `From<u64>` implementation :/

src/types/uint.rs
tomusdrw: Fixed!

src/types/uint.rs
rphmeier: Hm, I guess since `AsRef<[u8]>` _could_ be added to `u64` eventually :(
We'll have to wait for specialization then.

PR #31, opened 2017-05-24
Will you update the definition of Log?

usd-yamazaki:
I would like to use the Log address field.
Thank you.


usd-yamazaki:
The struct referred to this.
https://github.com/paritytech/parity/blob/master/rpc/src/v1/types/log.rs

tomusdrw:
Thanks!

PR #33, opened 2017-05-28
Various small improvements

tomusdrw:
- Expose `helpers` (makes it simpler to write custom APIs)
- Add `Deref<Target=[u8]>` to uints and `From<&[u8]>` for `Bytes`
- Make `Transport` blanket implementation more generic (covers `Arc` and `Rc`)

### DIFF
--- a/src/types/bytes.rs
+++ b/src/types/bytes.rs
@@ -7,9 +7,9 @@ use rustc_serialize::hex::{FromHex, ToHex};
 #[derive(Clone, Debug, Default, PartialEq)]
 pub struct Bytes(pub Vec<u8>);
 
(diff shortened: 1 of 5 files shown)


src/types/bytes.rs
```diff
@@ -7,6 +7,12 @@ use rustc_serialize::hex::{FromHex, ToHex};
 #[derive(Clone, Debug, Default, PartialEq)]
 pub struct Bytes(pub Vec<u8>);
 
+impl<'a> From<&'a [u8]> for Bytes {
```
rphmeier: could even be `From<T: Into<Vec<u8>>>`, right?

tomusdrw: Good point. Fixed.

PR #35, opened 2017-05-29
fixed filter names

debris:
Also made `CallResult` field public, cause it is used by many public functions

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -28,7 +28,6 @@ pub mod transports;
 pub mod types;
 
 pub use api::{Web3Main as Web3, ErasedWeb3};
-pub use helpers::CallResult;
 
 /// RPC result
 pub type Result<T> = futures::BoxFuture<T, Error>;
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -28,6 +28,7 @@ pub mod transports;
 pub mod types;
 
 pub use api::{Web3Main as Web3, ErasedWeb3};
+pub use helpers::CallResult;
```
tomusdrw: helpers are now public, this is unnecessary.

debris: imo it helps with docs readability -> https://tomusdrw.github.io/rust-web3/web3/api/eth/struct.Eth.html

All functions return `CallResult` and to know what this structure is, I need to open the source code

tomusdrw: Ah, sorry didn't notice that my PR is not merged yet.
This #33 exposes the whole `helpers` module, so docs will be fixed for `CallResult` as well. That's why I think that re-export is not necessary.

rphmeier: CallResult also isn't the best name...it should be called "LazyDecode" or something like that

rphmeier: (and `CallResult::new` everywhere isn't great either, maybe a `lazy_decode` helper function as a constructor would improve readability)

debris: @rphmeier I agree #38

PR #39, opened 2017-06-02
BaseFilter::stream

debris:
build on top of #35

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,6 +22,7 @@ serde_derive = "1.0"
 hyper = { version = "0.10", optional = true }
 tokio-core = { version = "0.1", optional = true }
 tokio-uds = { version = "0.1", optional = true }
+tokio-timer = "0.1"
 parking_lot = { version = "0.4", optional = true }
 
 [dev-dependencies]
@@ -34,3 +35,5 @@ default = ["http", "ipc"]
 # TODO [ToDr] move transports to separate crates
 http = ["hyper"]
 ipc = ["tokio-uds", "tokio-core", "parking_lot"]
+
+[workspace]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -10,14 +10,14 @@ keywords = ["jsonrpc", "web3", "ethereum", "rpc", "client"]
 authors = ["Tomasz Drwięga <tomasz@ethcore.io>"]
 
 [dependencies]
-ethabi = "1.0"
+ethabi = "1.1"
 futures = "0.1"
-jsonrpc-core = "6.0"
+jsonrpc-core = "7.0"
 log = "0.3"
 rustc-serialize = "0.3"
-serde = "0.9"
-serde_json = "0.9"
-serde_derive = "0.9"
+serde = "1.0"
+serde_json = "1.0"
+serde_derive = "1.0"
 # Optional deps
 hyper = { version = "0.10", optional = true }
 tokio-core = { version = "0.1", optional = true }
--- a/src/api/eth_filter.rs
+++ b/src/api/eth_filter.rs
@@ -36,7 +36,7 @@ impl<T: Transport + Clone> EthFilter<T> {
   pub fn new_blocks_filter(&self) -> FilterResult<T, CallResult<U256, T::Out>, BlocksFilter<T>> {
     FilterResult::new(
       self.transport.clone(),
-      CallResult::new(self.transport.execute("eth_newFilter", vec![]))
+      CallResult::new(self.transport.execute("eth_newBlockFilter", vec![]))
     )
   }
 
@@ -44,7 +44,7 @@ impl<T: Transport + Clone> EthFilter<T> {
   pub fn new_pending_transactions_filter(&self) -> FilterResult<T, CallResult<U256, T::Out>, PendingTransactionsFilter<T>> {
     FilterResult::new(
       self.transport.clone(),
-      CallResult::new(self.transport.execute("eth_newFilter", vec![]))
+      CallResult::new(self.transport.execute("eth_newPendingTransactionFilter", vec![]))
     )
   }
 }
@@ -333,7 +333,7 @@ mod tests {
     };
 
     // then
-    transport.assert_request("eth_newFilter", &[]);
+    transport.assert_request("eth_newBlockFilter", &[]);
     transport.assert_request("eth_uninstallFilter", &[r#""0x123""#.into()]);
     transport.assert_no_more_requests();
   }
@@ -357,7 +357,7 @@ mod tests {
 
     // then
     assert_eq!(result, Ok(Some(vec![0x456.into()])));
-    transport.assert_request("eth_newFilter", &[]);
+    transport.assert_request("eth_newBlockFilter", &[]);
     transport.assert_request("eth_getFilterChanges", &[r#""0x123""#.into()]);
     transport.assert_request("eth_uninstallFilter", &[r#""0x123""#.into()]);
     transport.assert_no_more_requests();
@@ -377,7 +377,7 @@ mod tests {
     };
 
     // then
-    transport.assert_request("eth_newFilter", &[]);
+    transport.assert_request("eth_newPendingTransactionFilter", &[]);
     transport.assert_request("eth_uninstallFilter", &[r#""0x123""#.into()]);
     transport.assert_no_more_requests();
   }
@@ -394,14 +394,14 @@ mod tests {
       let eth = EthFilter::new(&transport);
 
       // when
-      let filter = eth.new_blocks_filter().wait().unwrap();
+      let filter = eth.new_pending_transactions_filter().wait().unwrap();
       assert_eq!(filter.id, 0x123.into());
       filter.poll().wait()
     };
 
     // then
     assert_eq!(result, Ok(Some(vec![0x456.into()])));
-    transport.assert_request("eth_newFilter", &[]);
+    transport.assert_request("eth_newPendingTransactionFilter", &[]);
     transport.assert_request("eth_getFilterChanges", &[r#""0x123""#.into()]);
     transport.assert_request("eth_uninstallFilter", &[r#""0x123""#.into()]);
     transport.assert_no_more_requests();
--- a/src/api/eth_filter.rs
+++ b/src/api/eth_filter.rs
@@ -5,16 +5,19 @@ use std::marker::PhantomData;
 use std::time::Duration;
 use serde::de::DeserializeOwned;
 use tokio_timer::Timer;
-use futures::{Async, Poll, Future, Stream, stream};
+use futures::{Poll, Future, Stream, stream};
 
 use api::Namespace;
 use helpers::{self, CallResult};
 use types::{Filter, H256, Log, U256};
 use {Transport, Error, rpc};
 
+/// Specifies filter items and constructor method.
 pub trait FilterInterface {
+  /// Filter item type
   type Item;
 
+  /// Name of method used to construct the filter
   fn constructor() -> &'static str;
 }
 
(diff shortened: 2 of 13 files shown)


Cargo.toml
```diff
@@ -22,6 +22,7 @@ serde_derive = "1.0"
 hyper = { version = "0.10", optional = true }
 tokio-core = { version = "0.1", optional = true }
 tokio-uds = { version = "0.1", optional = true }
+tokio-timer = "0.1"
```
debris: it depends only on `futures` and `slab`

src/api/eth_filter.rs
```diff
@@ -103,6 +107,20 @@ impl<T: Transport, R> BaseFilter<T, R> {
...
   }
 
+  pub fn stream<'a>(&'a self, poll_interval: Duration) -> Box<Stream<Item = R, Error = Error> + 'a> where
+    T: Sync,
+    R: DeserializeOwned + Send + Sync + 'static,
+    T::Out: Send + 'static,
+  {
+    let result = Timer::default().interval(poll_interval)
+      .map_err(|_| Error::Unreachable)
+      .then(move |_| self.poll().map(|optional| optional.unwrap_or_else(Default::default)))
```
debris: map `Future<Item = Some<Vec<_>>, _> -> Future<Item = Vec<_>, _>`

src/api/eth_filter.rs
```diff
@@ -103,6 +107,20 @@ impl<T: Transport, R> BaseFilter<T, R> {
...
+      .map(|res| res.into_iter().map(Ok).collect::<Vec<Result<_, Error>>>())
+      .map(stream::iter)
+      .flatten();
```
debris: flatten it, so each event / block is a separate item of the stream

src/helpers.rs
```diff
@@ -85,23 +85,24 @@ pub mod tests {
...
   use rpc;
   use {Result, Error, Transport};
 
   #[derive(Default)]
   pub struct TestTransport {
     asserted: usize,
-    requests: RefCell<Vec<(String, Vec<rpc::Value>)>>,
-    response: RefCell<VecDeque<rpc::Value>>,
+    requests: RwLock<Vec<(String, Vec<rpc::Value>)>>,
+    response: RwLock<VecDeque<rpc::Value>>,
```
debris: When creating `BaseFilter::stream` i borrow `&BaseFilter`, which contains transport.

src/transports/ipc.rs
```diff
@@ -255,7 +255,7 @@ impl Future for WriteStream {
           }
         },
         WriteState::Writing { ref buffer, ref mut current_pos } => {
-          try_ready!(Ok(self.write.poll_write()));
+          try_ready!(Ok(self.write.poll_write()) as Result<_, ()>);
```
debris: after importing `tokio_timer` I had to explicitly specify Result error. Don't know why...

src/api/eth_filter.rs
```diff
@@ -103,6 +107,20 @@ impl<T: Transport, R> BaseFilter<T, R> {
     CallResult::new(self.transport.execute("eth_getFilterChanges", vec![id]))
   }
 
+  pub fn stream<'a>(&'a self, poll_interval: Duration) -> Box<Stream<Item = R, Error = Error> + 'a> where
```
tomusdrw: Wouldn't it be better to consume the filter? (then RwLocks in tests are not needed)

Also can we avoid `Boxing` the stream? By returning concrete struct that owns the filer we should be able to do that.

debris: Agree with consuming the filter! I will try prepare the struct, although it might not be trivial...

debris:
I altered the pr, so the filter is consumed when stream is created. I had to slightly redesign the architecture of filters to avoid `derefs`. Hope you like it :)

src/api/eth_filter.rs
```diff
@@ -1,108 +1,87 @@
...
+pub struct LogsFilter;
 
-  /// Installs a new block filter.
-  pub fn new_blocks_filter(&self) -> FilterResult<T, CallResult<U256, T::Out>, BlocksFilter<T>> {
-    FilterResult::new(
-      self.transport.clone(),
-      CallResult::new(self.transport.execute("eth_newFilter", vec![]))
-    )
-  }
+impl FilterInterface for LogsFilter {
```
rphmeier: why not just "Filter"? docs on this trait would be nice too

src/api/eth_filter.rs
```diff
@@ -114,104 +93,94 @@ impl<T: Transport, R> BaseFilter<T, R> {
...
-  fn new(id: U256, transport: T) -> Self {
-    BlocksFilter {
-      base: BaseFilter {
-        id: id,
-        transport: transport,
-        _output: PhantomData,
-      }
-    }
-  }
+pub struct CreateFilter<T: Transport, F: FilterInterface> {
```
rphmeier: docs? seems to be a future that resolves when the filter is created.

src/api/eth_filter.rs
```diff
@@ -114,104 +93,94 @@ impl<T: Transport, R> BaseFilter<T, R> {
...
+  type Item = BaseFilter<T, F>;
+  type Error = Error;
 
-  fn deref(&self) -> &Self::Target {
-    &self.base
+  fn poll(&mut self) -> Poll<Self::Item, Error> {
+    match self.future.poll() {
+      Ok(Async::Ready(x)) => Ok(Async::Ready(BaseFilter {
+        id: x,
+        transport: self.transport.clone(),
```
rphmeier: ready occurs only once, so we could remove the `Clone` bound by making `transport` an `Option<T>` and using `take`.

src/api/eth_filter.rs
```diff
@@ -114,104 +93,94 @@ impl<T: Transport, R> BaseFilter<T, R> {
...
+      transport: transport
     }
   }
 }
 
-impl<T: Transport> Deref for PendingTransactionsFilter<T> {
-  type Target = BaseFilter<T, H256>;
+impl<T: Transport + Clone> EthFilter<T> {
+  /// Installs a new logs filter.
+  pub fn new_logs_filter(&self, filter: Filter) -> CreateFilter<T, LogsFilter> {
```
rphmeier: what do you think of the naming convention `create_x_filter`?
or even just `logs()`? 

I'm thinking intended usage would be something like `my_web3.filter().logs()` which is a lot less repetitive.

src/api/eth_filter.rs
debris: there is already a structure called `Filter` in `src/types/logs.rs`

src/api/eth_filter.rs
debris: changed to `create_x_filter`

src/api/eth_filter.rs
```diff
@@ -114,104 +97,96 @@ impl<T: Transport, R> BaseFilter<T, R> {
...
-/// New blocks hashes filter.
-pub struct BlocksFilter<T: Transport> {
-  base: BaseFilter<T, H256>,
+/// Should be used to create new filter future
+pub fn create_filter<T: Transport, F: FilterInterface>(t: T, arg: Vec<rpc::Value>) -> CreateFilter<T, F> {
+  let future = CallResult::new(t.execute(F::constructor(), arg));
+  CreateFilter {
+    transport: Some(t),
+    interface: PhantomData,
+    future: future,
```
tomusdrw: Can use a shorthnd here: `future,`

src/api/eth_filter.rs
```diff
@@ -114,104 +97,96 @@ impl<T: Transport, R> BaseFilter<T, R> {
...
-        id: id,
-        transport: transport,
-        _output: PhantomData,
-      }
-    }
-  }
+/// Future which resolves with new filter
+pub struct CreateFilter<T: Transport, F: FilterInterface> {
+  transport: Option<T>,
+  interface: PhantomData<F>,
```
tomusdrw: It's not used so should be prefixed with `_`

src/api/eth_filter.rs
```diff
@@ -114,104 +97,96 @@ impl<T: Transport, R> BaseFilter<T, R> {
...
+  type Error = Error;
 
-  fn deref(&self) -> &Self::Target {
-    &self.base
+  fn poll(&mut self) -> Poll<Self::Item, Error> {
+    let id = try_ready!(self.future.poll());
+    let result = BaseFilter {
+      id: id,
+      transport: self.transport.take().expect("future polled after ready; qed"),
+      interface: PhantomData,
```
tomusdrw: Same here, `_interface`.

src/api/eth_filter.rs
```diff
@@ -114,104 +97,96 @@ impl<T: Transport, R> BaseFilter<T, R> {
...
+{
+  type Item = BaseFilter<T, F>;
+  type Error = Error;
 
-  fn deref(&self) -> &Self::Target {
-    &self.base
+  fn poll(&mut self) -> Poll<Self::Item, Error> {
+    let id = try_ready!(self.future.poll());
+    let result = BaseFilter {
+      id: id,
```
tomusdrw: Shorthand as well.

src/api/eth_filter.rs
```diff
@@ -114,104 +97,96 @@ impl<T: Transport, R> BaseFilter<T, R> {
...
-    BlocksFilter {
-      base: BaseFilter {
-        id: id,
-        transport: transport,
-        _output: PhantomData,
-      }
-    }
-  }
+/// Future which resolves with new filter
+pub struct CreateFilter<T: Transport, F: FilterInterface> {
```
tomusdrw: why carry around the whole `FilterInterface`? I think just `F::Item` would be enough.

src/api/eth_filter.rs
```diff
@@ -114,104 +97,96 @@ impl<T: Transport, R> BaseFilter<T, R> {
...
-    PendingTransactionsFilter {
-      base: BaseFilter {
-        id: id,
-        transport: transport,
-        _output: PhantomData,
-      }
+impl<T: Transport + Clone> Namespace<T> for EthFilter<T> {
+  fn new(transport: T) -> Self where Self: Sized {
+    EthFilter {
+      transport: transport
```
tomusdrw: shorthand.

src/api/eth_filter.rs
```diff
@@ -114,104 +97,96 @@ impl<T: Transport, R> BaseFilter<T, R> {
...
   }
 }
 
-impl<T: Transport> Deref for PendingTransactionsFilter<T> {
-  type Target = BaseFilter<T, H256>;
+impl<T: Transport + Clone> EthFilter<T> {
+  /// Installs a new logs filter.
+  pub fn create_logs_filter(&self, filter: Filter) -> CreateFilter<T, LogsFilter> {
+    let f = helpers::serialize(&filter);
+    create_filter(self.transport.clone(), vec![f])
```
tomusdrw: `Transport` doesn't need to be `Clone`, just do `&self.transport` here.

src/api/mod.rs
```diff
@@ -1,13 +1,12 @@
 //! `Web3` implementation
 
 mod eth;
-mod eth_filter;
+pub mod eth_filter;
```
tomusdrw: That's bad, better re-export everything for convenience. We don't really need to expose the module structure details imho.

src/transports/ipc.rs
tomusdrw: There is probably an `Into` implementation there that breaks things.

src/api/eth_filter.rs
```diff
@@ -114,104 +97,96 @@ impl<T: Transport, R> BaseFilter<T, R> {
...
+/// Future which resolves with new filter
+pub struct CreateFilter<T: Transport, F: FilterInterface> {
+  transport: Option<T>,
+  interface: PhantomData<F>,
+  future: CallResult<U256, T::Out>,
 }
 
-impl<T: Transport> Deref for BlocksFilter<T> {
-  type Target = BaseFilter<T, H256>;
+impl<T, F> Future for CreateFilter<T, F> where
```
tomusdrw: Just wondering if we really need a separate `Future` that resolves to `BaseFilter`. Can't the `BaseFilter` be stateful? Basically there is nothing you can do except for `and_then` on that filter, cause you need to wait for it to be created.

src/api/eth_filter.rs
debris: PhantomData does not have to be prefixed with `_`

PR #40, opened 2017-06-05
wait_for_confirmations

debris:


src/api/eth_filter.rs
```diff
@@ -1,108 +1,138 @@
...
+  let eth = EthFilter::new(transport);
+  let result = eth.create_blocks_filter()
+    .and_then(move |filter| {
+      filter.stream(Duration::from_secs(1))
+        .skip_while(move |hash| validate(hash).map(|ok| !ok))
+        .skip(1)
+        .take_while(validate)
+        .take(confirmations as u64)
+        .collect()
+        .and_then(move |hashes| if hashes.len() == confirmations {
```
debris: Currently, parity does not inform about forks happening. The next hash is always a hash of the `previous_block_number + 1`, even if there was a fork in the meantime. If event happened on both chains, let's not bother with possible difference in confirmation counter.

The different approach to the problem, would be polling `eth_blockNumber`, but it also has it's cons

src/api/eth_filter.rs
```diff
@@ -1,108 +1,138 @@
...
+pub fn wait_for_confirmations<'a, T, F, V>(transport: T, confirmations: usize, validate: &'a V)
+  -> Box<Future<Item = (), Error = Error> + 'a> where
+  T: 'a + Transport + Clone,
+  F: 'a + Future<Item = bool, Error = Error>,
+  V: 'a + Fn(&H256) -> F,
+{
+  let retries = 3;
+  let result = stream::repeat::<_, Error>(())
+    .take(retries)
+    .then(move |_| wait_for_confirmations_internal(transport.clone(), confirmations, validate))
```
debris: retry checking `x` consecutive blocks up to 3 times

src/confirm.rs
```diff
@@ -0,0 +1,116 @@
+use std::time::Duration;
+use futures::{Future, Stream};
+use api::{Eth, EthFilter, Namespace};
+use types::{H256, U256, TransactionRequest};
+use {Transport, Error};
+
+pub fn wait_for_confirmations<'a, T, F, V>(transport: T, poll_interval: Duration, confirmations: usize, check: V)
```
tomusdrw: I think it would be better if it was part of some kind of namespace.
The idea is that you're creating `Web3` object and you can do everything from there.
Now you would need to create two separate transports (one for Web3 and one for `wait_for_confirmations`) and use the function separately.

Maybe put it directly to `Web3Main`?

src/types/uint.rs
```diff
@@ -115,13 +115,23 @@ macro_rules! impl_uint {
...
         for i in 0..8 {
-          arr[$len - 1 - i] =  num as u8;
+          arr[$len - 1 - i] = num as u8;
           num = num >> 8;
         }
         $name(arr)
       }
     }
 
+    impl From<$name> for u64 {
```
tomusdrw: This kind of implies that the conversion is lossless, I would rather have `low_u64()` as we have in bigint.

src/confirm.rs
```diff
@@ -0,0 +1,116 @@
...
+  -> Box<Future<Item = (), Error = Error> + 'a> where
+  T: 'a + Transport + Clone,
+  F: 'a + Future<Item = Option<U256>, Error = Error>,
```
rphmeier: `IntoFuture` is more general, then you can just return `Result` from that function

src/api/mod.rs
```diff
@@ -67,4 +70,21 @@ impl<T: Transport> Web3Main<T> {
...
+  -> Box<Future<Item = (), Error = Error> + 'a> where
+  F: 'a + IntoFuture<Item = Option<U256>, Error = Error>,
+  V: 'a + Fn() -> F,
+  {
+    confirm::wait_for_confirmations(&self.transport, poll_interval, confirmations, check)
+  }
+
+
+  /// Sends transaction and returns future resolved after transaction is confirmed
+  pub fn send_transaction_with_confirmation<'a>(&'a self, tx: TransactionRequest, poll_interval: Duration, confirmations: usize)
```
tomusdrw: IMHO this fits more to `Eth` namespace, but no strong feelings.

src/confirm.rs
```diff
@@ -0,0 +1,129 @@
...
+  T: 'a + Transport + Clone {
+  let eth = Eth::new(transport.clone());
+  let result = eth.send_transaction(tx)
+    .and_then(move |hash| {
+      wait_for_confirmations(transport.clone(), poll_interval, confirmations, move || {
+        eth.transaction_receipt(hash.clone()).map(|option| option.map(|receipt| receipt.block_number))
+      })
+      .and_then(move |_| {
+        let eth = Eth::new(transport.clone());
+        eth.transaction_receipt(hash).and_then(|option| match option {
```
tomusdrw: `option.ok_or(Error::Unreachable)`?

PR #41, opened 2017-06-05
Initial version of batch requests.

tomusdrw:
(Still in progress (need to get rid of Box), but would like you guys to have a look already).

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -7,10 +7,9 @@ extern crate jsonrpc_core as rpc;
 extern crate parking_lot;
 extern crate rustc_serialize;
 extern crate serde;
+extern crate serde_json;
 extern crate tokio_timer;
 
-#[macro_use]
-extern crate serde_json;
 #[macro_use]
 extern crate log;
 #[macro_use]
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -7,8 +7,10 @@ extern crate jsonrpc_core as rpc;
 extern crate parking_lot;
 extern crate rustc_serialize;
 extern crate serde;
-extern crate serde_json;
 extern crate tokio_timer;
+
+#[cfg_attr(test, macro_use)]
+extern crate serde_json;
 
 #[macro_use]
 extern crate log;
@@ -19,13 +21,15 @@ extern crate serde_derive;
 #[macro_use]
 pub extern crate futures;
 
+// it needs to be before other modules
+// otherwise the macro for tests is not available.
+#[macro_use]
+pub mod helpers;
+
 pub mod api;
 pub mod contract;
 pub mod transports;
 pub mod types;
-
-#[macro_use]
-pub mod helpers;
 
 mod confirm;
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -62,6 +62,7 @@ impl From<rpc::Error> for Error {
   }
 }
 
+/// Assigned RequestId
 pub type RequestId = usize;
 
 /// Transport implementation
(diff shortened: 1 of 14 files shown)


src/lib.rs
```diff
@@ -104,8 +140,23 @@ impl<X, T> Transport for X where
...
+}
+
+impl<X, T> BatchTransport for X where
+  T: BatchTransport + ?Sized,
+  X: ::std::ops::Deref<Target=T>,
+{
+  type Batch = T::Batch;
+
+  fn send_batch(&self, requests: Vec<(RequestId, rpc::Call)>) -> Self::Batch {
+    (**self).send_batch(requests)
```
debris: is it really needed? I mean:

```rust
+impl<X, T> BatchTransport for X where
 +  T: BatchTransport + ?Sized,
```

src/transports/ipc.rs
```diff
@@ -60,22 +60,24 @@ impl Drop for EventLoopHandle {
 }
 
 
-/// Error returned while initializing IPC transport
+/// Error returned while initializing IPC transport.
 pub type IpcError = io::Error;
+/// Result of initializing IPC transport.
+pub type IpcResult<T> = ::std::result::Result<T, IpcError>;
```
rphmeier: Style is typically to declare `type Result` and then it can be used as `ipc::Result`

tomusdrw:
Ping: @debris @rphmeier

src/transports/batch.rs
```diff
@@ -0,0 +1,98 @@
...
+use transports::Result;
+use {BatchTransport, Transport, Error as RpcError, RequestId};
+
+type Pending = oneshot::Sender<Result<rpc::Value>>;
+
+/// Transport allowing to batch queries together.
+pub struct Batch<T> {
+  transport: T,
+  pending: Arc<Mutex<BTreeMap<RequestId, Pending>>>,
+  batch: Mutex<Vec<(RequestId, rpc::Call)>>,
```
rphmeier: could probably kill `pending` field and put it inside the mutex here as far as I can tell each batch should have its own pending set.

rphmeier: i guess it's a slight optimization that you don't have to reallocate the pending set, but it also means that queuing a new batch will be blocked while processing the response of the last one.

src/transports/ipc.rs
```diff
@@ -156,44 +157,91 @@ impl Ipc {
...
+    let (tx, rx) = futures::oneshot();
+    self.pending.lock().insert(id, tx);
+    let result = {
+      let mut sender = self.write_sender.lock();
+      (*sender).send(request.into_bytes())
+    };
 
     IpcTask {
       state: IpcTaskState::Sending(Some(result), rx),
+      extract: batch_response as fn(Vec<RpcResult<rpc::Value>>) -> RpcResult<Vec<RpcResult<rpc::Value>>>,
```
rphmeier: you don't technically need a separate `batch_response` function for this: enum variants are their own functions so I think `Ok as fn(...) -> ...` would work

rphmeier:
Passing a `Vec<(RequestId, Call)>` to `send_batch` when the implementations just take out the calls (and allocate again) doesn't seem quite right. Passing an `I: IntoIterator<Item=(RequestId, Call)>` would be a lot more flexible. But I guess it would bar any future implementation of `BatchTransport` for type-erased transport until specialization is stable.

Also a utility method on `Transport`: `fn batched(self) -> Batch<Self> where Self: Sized + BatchTransport` would be cool

tomusdrw:
Updated.
- [x] `send_batch` now takes `IntoIterator` instead of `Vec`
- [x] `submit_batch` returns concrete type instead of `BoxFuture`.

PR #42, opened 2017-06-13
Fix backpressure in IPC transport.

tomusdrw:
Cloning a channel creates a buffer slot for it which could result in unbounded queue if the requests are produced faster than they can be processed.


debris:
lgtm, although it's a shame that it makes everything sync. But I can't see any other solution to the problem :/

PR #46, opened 2017-07-10
removed lifetimes 

debris:
had difficulties using `send_transaction_with_confirmation` so I removed lifetimes from `EthFilter::stream`, `wait_for_confirmations` and `send_transaction_with_confirmation`

### DIFF
--- a/src/api/mod.rs
+++ b/src/api/mod.rs
@@ -13,9 +13,9 @@ pub use self::personal::Personal;
 pub use self::web3::Web3;
 
 use std::time::Duration;
-use futures::{Future, IntoFuture};
+use futures::{IntoFuture};
 use {confirm, Transport, Error};
-use types::{U256, TransactionRequest, TransactionReceipt};
+use types::{U256, TransactionRequest};
 
 /// Common API for all namespaces
 pub trait Namespace<T: Transport> {
@@ -80,17 +80,25 @@ impl<T: Transport> Web3Main<T> {
   }
 
   /// Should be used to wait for confirmations
-  pub fn wait_for_confirmations<F, V>(&self, poll_interval: Duration, confirmations: u64, check: V)
-    -> confirm::Confirmations<&T, V, F::Future> where
-  F: IntoFuture<Item = Option<U256>, Error = Error>,
-  V: confirm::ConfirmationCheck<Check = F>
+  pub fn wait_for_confirmations<F, V>(
+    &self,
+    poll_interval: Duration,
+    confirmations: u64,
+    check: V
+  ) -> confirm::Confirmations<&T, V, F::Future> where
+    F: IntoFuture<Item = Option<U256>, Error = Error>,
+    V: confirm::ConfirmationCheck<Check = F>,
   {
     confirm::wait_for_confirmations(&self.transport, poll_interval, confirmations, check)
   }
 
   /// Sends transaction and returns future resolved after transaction is confirmed
-  pub fn send_transaction_with_confirmation(&self, tx: TransactionRequest, poll_interval: Duration, confirmations: u64)
-  -> confirm::SendTransactionWithConfirmation<&T> {
+  pub fn send_transaction_with_confirmation(
+    &self,
+    tx: TransactionRequest,
+    poll_interval: Duration,
+    confirmations: u64
+  ) -> confirm::SendTransactionWithConfirmation<&T> {
     confirm::send_transaction_with_confirmation(&self.transport, tx, poll_interval, confirmations)
   }
 }
--- a/src/api/mod.rs
+++ b/src/api/mod.rs
@@ -7,7 +7,7 @@ mod personal;
 mod web3;
 
 pub use self::eth::Eth;
-pub use self::eth_filter::{BaseFilter, CreateFilter, EthFilter};
+pub use self::eth_filter::{BaseFilter, CreateFilter, EthFilter, FilterStream};
 pub use self::net::Net;
 pub use self::personal::Personal;
 pub use self::web3::Web3;
@@ -80,19 +80,17 @@ impl<T: Transport> Web3Main<T> {
   }
 
   /// Should be used to wait for confirmations
-  pub fn wait_for_confirmations<'a, F, V>(&'a self, poll_interval: Duration, confirmations: usize, check: V)
-  -> Box<Future<Item = (), Error = Error> + 'a> where
-  F: 'a + IntoFuture<Item = Option<U256>, Error = Error>,
-  V: 'a + Fn() -> F,
+  pub fn wait_for_confirmations<F, V>(&self, poll_interval: Duration, confirmations: u64, check: V)
+    -> confirm::Confirmations<&T, V, F::Future> where
+  F: IntoFuture<Item = Option<U256>, Error = Error>,
+  V: confirm::ConfirmationCheck<Check = F>
   {
     confirm::wait_for_confirmations(&self.transport, poll_interval, confirmations, check)
   }
 
-
   /// Sends transaction and returns future resolved after transaction is confirmed
-  pub fn send_transaction_with_confirmation<'a>(&'a self, tx: TransactionRequest, poll_interval: Duration, confirmations: usize)
-  -> Box<Future<Item = TransactionReceipt, Error = Error> + 'a> where
-  {
+  pub fn send_transaction_with_confirmation(&self, tx: TransactionRequest, poll_interval: Duration, confirmations: u64)
+  -> confirm::SendTransactionWithConfirmation<&T> {
     confirm::send_transaction_with_confirmation(&self.transport, tx, poll_interval, confirmations)
   }
 }
--- a/src/confirm.rs
+++ b/src/confirm.rs
@@ -1,5 +1,5 @@
 use std::time::Duration;
-use futures::{IntoFuture, Future, Stream, Poll, Async};
+use futures::{IntoFuture, Future, Stream, Poll};
 use futures::stream::Skip;
 use api::{Eth, EthFilter, Namespace, CreateFilter, FilterStream};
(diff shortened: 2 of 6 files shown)


src/api/mod.rs
```diff
@@ -80,19 +80,17 @@ impl<T: Transport> Web3Main<T> {
   }
 
   /// Should be used to wait for confirmations
-  pub fn wait_for_confirmations<'a, F, V>(&'a self, poll_interval: Duration, confirmations: usize, check: V)
-  -> Box<Future<Item = (), Error = Error> + 'a> where
-  F: 'a + IntoFuture<Item = Option<U256>, Error = Error>,
-  V: 'a + Fn() -> F,
+  pub fn wait_for_confirmations<F, V>(&self, poll_interval: Duration, confirmations: u64, check: V)
+    -> confirm::Confirmations<&T, V, F::Future> where
```
tomusdrw: Style nitpick: I would rather see the output type on the same line as the function name and type parameters idented.
If it doesn't fit then it should be split by parameters:
```rust
pub fn wait_for_confirmations<F, V>(
  &self,
  poll_interval: Duration,
  confirmations: u64,
  check: V,
) -> confirm::Confirmations<&T, V, F::Future> where
  F: IntoFuture<..>,
  V: confirm::ConfirmationCheck<Check = F>,
{
 ...
}
```

src/confirm.rs
```diff
@@ -1,48 +1,230 @@
...
-  let eth = Eth::new(transport.clone());
-  let result = eth.send_transaction(tx)
-    .and_then(move |hash| {
-      wait_for_confirmations(transport.clone(), poll_interval, confirmations, move || {
-        eth.transaction_receipt(hash.clone()).map(|option| option.map(|receipt| receipt.block_number))
-      })
-      .and_then(move |_| {
-        let eth = Eth::new(transport.clone());
-        eth.transaction_receipt(hash).and_then(|option| option.ok_or(Error::Unreachable))
+pub trait ConfirmationCheck {
```
tomusdrw: Should it have a blanket implementation for `T, F where F: Fn() -> T, T: IntoFuture<Item=Option<U256>, Error>`?

src/confirm.rs
```diff
@@ -1,48 +1,230 @@
...
+  state: WaitForConfirmationsState<F, T::Out>,
+  filter_stream: Skip<FilterStream<T, H256>>,
+  confirmation_check: V,
+  confirmations: u64,
+}
+
+impl<T, V, F> Future for WaitForConfirmations<T, V, F::Future> where
+  T: Transport,
+  V: ConfirmationCheck<Check = F>,
+  F: IntoFuture<Item = Option<U256>, Error = Error> {
```
tomusdrw: Style nitpick: missing trailing `,` and `{` should be on a new line.

src/confirm.rs
```diff
@@ -1,48 +1,230 @@
...
+  fn poll(&mut self) -> Poll<Self::Item, Self::Error> {
+    loop {
+      let next_state = match self.state {
+        WaitForConfirmationsState::WaitForNextBlock => {
+          let _ = try_ready!(self.filter_stream.poll());
+          WaitForConfirmationsState::CheckConfirmation(self.confirmation_check.check().into_future())
+        },
+        WaitForConfirmationsState::CheckConfirmation(ref mut future) => match try_ready!(future.poll()) {
+          Some(confirmation_block_number) => {
+            let future = Eth::new(&self.transport).block_number();
```
tomusdrw: Can't we do it in parallel with `confirmation_check.check()`?

src/confirm.rs
```diff
@@ -1,48 +1,230 @@
...
+  Wait(WaitForConfirmations<T, V, F>),
+}
+
+pub struct Confirmations<T: Transport, V, F> {
+  state: ConfirmationsState<T, V, F>,
+}
+
+impl<T: Transport + Clone, V, F> Confirmations<T, V, F> {
+  fn new(transport: T, poll_interval: Duration, confirmations: u64, check: V) -> Self {
+    let eth = EthFilter::new(transport.clone());
```
tomusdrw: I think this could be just `&transport` here (and `+Clone` bound would be redundant)

src/confirm.rs
```diff
@@ -1,48 +1,230 @@
...
-    });
-  Box::new(result)
+    }
+  }
+}
+
+impl<T, V, F> Future for Confirmations<T, V, F::Future> where
+  T: Transport,
+  V: ConfirmationCheck<Check = F>,
+  F: IntoFuture<Item = Option<U256>, Error = Error> {
```
tomusdrw: style nitpick (comma and separate line)

src/confirm.rs
```diff
@@ -1,48 +1,230 @@
...
+      };
+      self.state = next_state;
+    }
+  }
+}
+
+struct CreateWaitForConfirmations<T: Transport, V> {
+  create_filter: CreateFilter<T, H256>,
+  poll_interval: Duration,
+  transport: Option<T>,
```
tomusdrw: We `.take()` both of them, so maybe better store it as `Option<(T, V)>` to have only one `expect()`?

src/confirm.rs
```diff
@@ -1,48 +1,230 @@
...
+      };
+      self.state = next_state;
+    }
+  }
+}
+
+pub fn wait_for_confirmations<T, V, F>(transport: T, poll_interval: Duration, confirmations: u64, check: V) -> Confirmations<T, V, F::Future> where
+  T: Transport + Clone,
+  V: ConfirmationCheck<Check = F>,
+  F: IntoFuture<Item = Option<U256>, Error = Error>, {
```
tomusdrw: style nitpick - `{` in a separate line

src/confirm.rs
```diff
@@ -1,48 +1,230 @@
...
+  state: SendTransactionWithConfirmationState<T>,
+  eth: Eth<T>,
+  transport: T,
+  poll_interval: Duration,
+  confirmations: u64,
+}
+
+impl<T: Transport + Clone> SendTransactionWithConfirmation<T> {
+  fn new(transport: T, tx: TransactionRequest, poll_interval: Duration, confirmations: u64) -> Self {
+  let eth = Eth::new(transport.clone());
```
tomusdrw: Bad indentation.

src/confirm.rs
```diff
@@ -1,48 +1,230 @@
...
+  type Item = TransactionReceipt;
+  type Error = Error;
+
+  fn poll(&mut self) -> Poll<Self::Item, Self::Error> {
+    loop {
+      let next_state = match self.state {
+        SendTransactionWithConfirmationState::SendTransaction(ref mut future) => {
+          let hash = try_ready!(future.poll());
+          let confirmation_check = TransactionReceiptBlockNumberCheck::new(Eth::new(self.transport.clone()), hash.clone());
+          let wait = wait_for_confirmations(self.transport.clone(), self.poll_interval, self.confirmations, confirmation_check);
```
tomusdrw: Isn't it enough to pass `&self.transport`?

src/lib.rs
```diff
@@ -75,7 +75,7 @@ pub type RequestId = usize;
 /// Transport implementation
 pub trait Transport {
   /// The type of future this transport returns when a call is made.
-  type Out: futures::Future<Item=rpc::Value, Error=Error>;
+  type Out: futures::Future<Item=rpc::Value, Error=Error> + 'static;
```
tomusdrw: I think that's not really needed (remember we were just trying this out).

src/confirm.rs
debris: imo, there is no need to right now

src/confirm.rs
debris: I can't do that without using lifetimes, cause then `Confirmations` would contain 2 transports: `T` && `&T`.

src/confirm.rs
debris: same as above

debris: same as above

src/confirm.rs
debris: we can, but then we make redundant calls to get block number when `confirmation_check.check()` returns `None`

src/confirm.rs
debris: I agree, but in the same time, putting them in a tuple will be less descriptive

src/lib.rs
```diff
@@ -31,7 +31,7 @@ pub mod contract;
 pub mod transports;
 pub mod types;
 
-mod confirm;
+pub mod confirm;
```
tomusdrw: If it's public it should be aligned with other modules.

src/confirm.rs
```diff
@@ -1,48 +1,244 @@
...
+use types::{H256, U256, TransactionRequest, TransactionReceipt};
+use helpers::CallResult;
 use {Transport, Error};
 
-pub fn wait_for_confirmations<'a, T, F, V>(transport: T, poll_interval: Duration, confirmations: usize, check: V)
-  -> Box<Future<Item = (), Error = Error> + 'a> where
-  T: 'a + Transport + Clone,
-  F: 'a + IntoFuture<Item = Option<U256>, Error = Error>,
-  V: 'a + Fn() -> F,
+pub trait ConfirmationCheck {
```
tomusdrw: Since the module is now public all public items require a documentation.

src/types/transaction.rs
```diff
@@ -34,10 +34,6 @@ pub struct Transaction {
 /// "Receipt" of an executed transaction: details of its execution.
 #[derive(Debug, Default, Clone, PartialEq, Serialize, Deserialize)]
 pub struct Receipt {
-  /// Hash
-  pub hash: H256,
```
tomusdrw: Should be in a separate PR

debris:
of course :)

PR #48, opened 2017-08-03
Deserialize arrays and vectors.

### LINKED ISSUE #47: How do I get multiple return values from the contract function?
There is a function of such a contract.
`function getValues() public constant returns (uint[], byte[64][], address[])`

I wrote the Rust code to receive the value, but this is an error.
```
let result = contract.query("getValues", (), None, Options::default());
let values: Vec<Token> = result.wait().unwrap();
```
```
error[E0277]: the trait bound `std::vec::Vec<ethabi::Token>: web3::contract::tokens::Tokenizable` is not satisfied
```

It worked when you made changes to crate, but I think this is a bad way.

`use ethabi::Token;` -> `use web3::contract::tokens::Token;`

> rust-web3/src/contract/tokens.rs:3
`use ethabi::Token;` -> `pub use ethabi::Token;`

> ethabi/src/token/token.rs:7
`#[derive(Debug, PartialEq, Clone)]` -> `#[derive(Debug, PartialEq, Clone, Deserialize)]`

How is the best way to do it?
Thank you.

tomusdrw:
Closes #47

### DIFF
--- a/src/contract/tokens.rs
+++ b/src/contract/tokens.rs
@@ -6,12 +6,12 @@ use contract::Error;
 use types::{self, Address, H256, U256};
 
 /// Output type possible to deserialize from Contract ABI
-pub trait Output {
+pub trait Deserialize {
   /// Creates a new instance from parsed ABI tokens.
   fn from_tokens(tokens: Vec<Token>) -> Result<Self, Error> where Self: Sized;
 }
 
-impl<T: Tokenizable> Output for T {
+impl<T: Tokenizable> Deserialize for T {
   fn from_tokens(mut tokens: Vec<Token>) -> Result<Self, Error> {
     if tokens.len() != 1 {
       return Err(Error::InvalidOutputType(format!("Expected single element, got a list: {:?}", tokens)));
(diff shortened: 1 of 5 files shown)


src/contract/tokens.rs
```diff
@@ -1,42 +1,76 @@
...
+  fn from_tokens(mut tokens: Vec<Token>) -> Result<Self, Error> {
+    if tokens.len() != 1 {
+      return Err(Error::InvalidOutputType(format!("Expected single element, got a list: {:?}", tokens)));
+    }
+    Self::from_token(tokens.drain(..).next().expect("At least one element in vector; qed"))
+  }
+}
+
+macro_rules! impl_output {
+  ($num: expr, $( $ty: ident : $no: tt, )+) => {
```
rphmeier: is $no used anymore?

tomusdrw: Good catch. Updated.

tomusdrw:
What about `Detokenize` and `Tokenize`?

debris:
sounds about right

PR #49, opened 2017-08-03
fixed TransactionReceipt fields && added tests for deserialization

debris:
transaction receipt does not have fields called `index` and `hash` [docs](https://github.com/ethereum/wiki/wiki/JSON-RPC#eth_gettransactionreceipt)

### DIFF
--- a/src/confirm.rs
+++ b/src/confirm.rs
@@ -69,8 +69,8 @@ mod tests {
       condition: None,
     };
     let transaction_receipt = TransactionReceipt {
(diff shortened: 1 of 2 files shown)


tomusdrw:
Yes they do: https://github.com/paritytech/parity/wiki/JSONRPC-eth-module#eth_gettransactionreceipt
But the fields should be named differently.

(iirc it was even request by go-eth team)

debris:
you are absolutely right

PR #52, opened 2017-08-08
Change the return type used for eth_sign from H512 to H520.

usd-yamazaki:
Since the signature data consists of "r (32 byte)", "s (32 byte)" and "v (1 byte)", I think that it is correct to receive at H520 instead of H512.


tomusdrw:
You're right, it should be `H520`. Thank you!

PR #55, opened 2017-08-16
Rewrite http transport to use latest hyper.

### LINKED ISSUE #27: Use latest (async) hyper for HTTP transport

tomusdrw:
Resolves #27

### DIFF
--- a/src/transports/shared.rs
+++ b/src/transports/shared.rs
@@ -1,0 +1,138 @@
+use std::{thread, fmt, mem};
+use std::sync::{self, atomic, Arc};
+use futures::{self, Future};
+use futures::sync::oneshot;
+use transports::Result;
+use transports::tokio_core::reactor;
+use {Error, RequestId};
+
+/// Event Loop Handle.
+/// NOTE: Event loop is stopped when handle is dropped!
+pub struct EventLoopHandle {
+  thread: Option<thread::JoinHandle<()>>,
+  remote: reactor::Remote,
+  done: Arc<atomic::AtomicBool>,
+}
+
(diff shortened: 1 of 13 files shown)


src/transports/shared.rs
```diff
@@ -0,0 +1,139 @@
...
+      };
+
+      let res = run();
+      match res {
+        Err(e) => send(Err(e)),
+        Ok((http, mut event_loop)) => {
+          send(Ok((http, event_loop.remote())));
+
+          while !done2.load(atomic::Ordering::Relaxed) {
+            event_loop.turn(None);
```
rphmeier: does this work? won't it end up blocking forever instead of continuing with the loop?

tomusdrw: Yup, the docs says that infinite loop with `turn` is equivalent to calling `run` with a never ending future:
https://docs.rs/tokio-core/0.1.9/tokio_core/reactor/struct.Core.html#method.turn

rphmeier: but that isn't what we want, is it? We would prefer for the thread to join when the flag is set.

tomusdrw: AFAICT there is no way currently to figure out how many pending futures is there in the event loop to run all of them to completion.
Actually imho when one drops the EventLoopHandle (which causes this flag to be set) she doesn't really care about the futures that are still running.
Othere option would be to use a `oneshot` future here and `event_loop.run(oneshot)` that future, but it will lead to the same results - when the future is completed event loop will be stopped regardless of any other futures still running.

What solution do you have in mind?

PR #59, opened 2017-08-25
Adding contract deployment.

### LINKED ISSUE #53: Example showing how to deploy a new contract instance
Assuming that this is possible, an example showing how to deploy an instance of a new contract would be great. The example could assume that "solc" is installed on the system and shell out to it to compile the contract.

tomusdrw:
Closes #53


tomusdrw:
@debris ping

PR #64, opened 2017-09-21
Added BlockNumber parameter to Contract query

usd-yamazaki:
I want to add a parameter to call constant function by specifying a block number.
Thank you.

### DIFF
--- a/src/contract/mod.rs
+++ b/src/contract/mod.rs
@@ -126,9 +126,10 @@ impl<T: Transport> Contract<T> {
   }
 
   /// Call constant function
-  pub fn query<R, A, P>(&self, func: &str, params: P, from: A, options: Options, block: Option<BlockNumber>) -> QueryResult<R, T::Out> where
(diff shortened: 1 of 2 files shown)


src/contract/mod.rs
```diff
@@ -126,7 +126,7 @@ impl<T: Transport> Contract<T> {
   }
 
   /// Call constant function
-  pub fn query<R, A, P>(&self, func: &str, params: P, from: A, options: Options) -> QueryResult<R, T::Out> where
+  pub fn query<R, A, P>(&self, func: &str, params: P, from: A, options: Options, block: Option<BlockNumber>) -> QueryResult<R, T::Out> where
```
tomusdrw: Can you make it `B: Into<Option<BlockNumber>>`? This allows to omit `Some` when calling the function.

usd-yamazaki: Thank you.
I learned a lot!

usd-yamazaki:
I do not know why this test was canceled.

tomusdrw:
Me neither, but went well after restart. Thanks!

PR #68, opened 2017-10-01
Rewrite with Error Chain

### LINKED ISSUE #67: Use error_chain for errors.

tomusdrw:
Closes #67

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,6 +21,7 @@ serde = "1.0"
 serde_json = "1.0"
 serde_derive = "1.0"
 tokio-timer = "0.1"
+error-chain = "0.11.0-rc.2"
 # Optional deps
 hyper = { version = "0.11", optional = true }
 tokio-core = { version = "0.1", optional = true }
--- a/src/contract/error.rs
+++ b/src/contract/error.rs
@@ -1,0 +1,38 @@
+//! Contract call/query error.
+
+#![allow(unknown_lints)]
+#![allow(missing_docs)]
+
+use ethabi;
+
+error_chain! {
+  links {
+    Abi(ethabi::Error, ethabi::ErrorKind);
+    Api(::Error, ::ErrorKind);
+  }
+
+  errors {
+    InvalidOutputType(e: String) {
+      description("invalid output type requested by the caller"),
+      display("Invalid output type: {}", e),
+    }
+  }
+}
+
+/// Contract deployment error.
+pub mod deploy {
+  use types::H256;
+
+  error_chain! {
+    links {
+      Api(::Error, ::ErrorKind);
+    }
+
+    errors {
+      ContractDeploymentFailure(hash: H256) {
+        description("Contract deployment failed")
+        display("Failure during deployment. Tx hash: {:?}", hash),
+      }
+    }
+  }
+}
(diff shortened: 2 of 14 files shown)


Cargo.toml
```diff
@@ -21,6 +21,7 @@ serde = "1.0"
 serde_json = "1.0"
 serde_derive = "1.0"
 tokio-timer = "0.1"
+error-chain = "0.11.0-rc.2"
```
debris: please use `0.11`

src/contract/error.rs
```diff
@@ -0,0 +1,38 @@
...
+
+#![allow(unknown_lints)]
+#![allow(missing_docs)]
+
+use ethabi;
+
+error_chain! {
+  links {
+    Abi(ethabi::Error, ethabi::ErrorKind);
+    Api(::Error, ::ErrorKind);
```
debris: isn't it better to use `foreign_links` here?

src/error.rs
```diff
@@ -0,0 +1,88 @@
...
+    Internal {
+      description("web3 internal error"),
+      display("Internal Web3 error"),
+    }
+  }
+}
+
+impl From<serde_json::Error> for Error {
+  fn from(err: serde_json::Error) -> Self {
+    ErrorKind::Decoder(format!("{:?}", err)).into()
```
debris: can't we move it to the `foreign_links`?

src/error.rs
```diff
@@ -0,0 +1,88 @@
...
+  }
+}
+
+impl Clone for Error {
+  fn clone(&self) -> Self {
```
debris: why is it needed?

Cargo.toml
tomusdrw: Requires https://github.com/paritytech/ethabi/issues/49

src/contract/error.rs
tomusdrw: `links` are used for other `error_chains`.

src/error.rs
tomusdrw: Let me check again, but I think `serde_json::Error` does not implement `Error` trait (which is required)

src/error.rs
tomusdrw: It's needed for batch requests, where you can do:
`let accounts = web3.eth().accounts();`
`let batch = web3.transport().send_batch()`;
and both futures need to resolve to the same value.

src/error.rs
tomusdrw: The problem is with `Clone` actually, `serde_json::Error` can't be cloned and there is no way to re-create the error.

src/error.rs
```diff
@@ -0,0 +1,88 @@
+//! Web3 Error
+
+#![allow(unknown_lints)]
+#![allow(missing_docs)]
+
+use std::io;
+use serde_json;
+use rpc;
+
+error_chain! {
```
vorner: Hello

Thank you for the change :-). Though I might clarify that I asked for something slightly different ‒ I wanted your errors to implement the `std::error::Error` trait, which allows it to be used by a `foreign_link` inside `error_chain` on the user side (or, in general, to act as well-behaved error). But this works too (because `error_chain` implements the trait).

tomusdrw: Hey! Yeah, I know that `Error` trait would be enough, but error chain is awesome and we are migrating all our libraries to it anyways:)

PR #73, opened 2017-10-12
Fixed that CLOSE_WAIT connection remains for a while.

usd-yamazaki:
When accessing Parity, the problem that CLOSE_WAIT state connection remains for a while (about 2 minutes) is fixed.
This problem becomes more noticeable when access to Parity is repeated.

### DIFF
--- a/src/transports/http.rs
+++ b/src/transports/http.rs
@@ -110,6 +110,7 @@ impl Http {
     let mut req = hyper::client::Request::new(hyper::Method::Post, self.url.clone());
     req.headers_mut().set(hyper::header::ContentType::json());
     req.headers_mut().set(hyper::header::UserAgent::new("web3.rs"));
+    req.headers_mut().set(hyper::header::Connection::close());
     let len = request.len();
     // Don't send chunked request
     if len < MAX_SINGLE_CHUNK {


usd-yamazaki:
I am making a server program using jsonrpc-http-server. (Thanks as always!)

When there is a request from the client, the server accesses the parity and returns the retrieved value.
For example, such a function is called.

```
fn get_nonce(&self, address: Address) -> Result<U256, Error> {
    let (_eloop, transport) = Http::new(&super::PARITY_ADDRESS).unwrap();
    let rpc = Web3::new(&transport);
    let nonce = rpc.eth().transaction_count(address, None).wait().unwrap();
    Ok(nonce)
}
```

When requests from clients increase, the connections of CLOSE_WAIT will remain in large quantities and eventually will not be able to connect to parity.

Connection state image. (8021 is the port of parity.)
https://i.imgur.com/VuxeknC.png

tomusdrw:
I see. Your currrent approach is pretty resource-heavy. You are creating a new thread with EventLoop and new transport with connection pool for each call of `get_nonce`. You should rather re-use existing transport:
```rust
SomeStruct {
 rpc: Web3<Http>, 
}

fn get_nonce(&self, address: Address) -> Result<U256, Error> {
//    let (_eloop, transport) = Http::new(&super::PARITY_ADDRESS).unwrap();
//    let rpc = Web3::new(&transport);
    let nonce = self.rpc.eth().transaction_count(address, None).wait().unwrap();
    Ok(nonce)
}
```

That said the option to have `Connection: close` would be really good to have.
I also now started wondering if we shouldn't force connection termination in `Drop` implementation of http transport (my guess is that right now because event loop is dropped as well, they are just timing-out).

usd-yamazaki:
Thanks for the advice.
After correcting like this, the problem was solved.
I will withdraw the pull request for now.

```
pub struct SomeStruct {
    _eloop: EventLoopHandle,
    rpc: Web3<Http>,
}

impl Server for SomeStruct {
    fn get_nonce(&self, address: Address) -> Result<U256, Error> {
        let nonce = self.rpc.eth().transaction_count(address, None).wait().unwrap();
        Ok(nonce)
    }
}

impl SomeStruct {
    pub fn new() -> Result<SomeStruct, Web3Error> {
        let (eloop, transport) = Http::new(&super::PARITY_ADDRESS)?;
        Ok(SomeStruct {
            _eloop: eloop,
            rpc: Web3::new(transport),
        })
    }
}
```

Issue #16, opened 2017-01-03
Move transports to separate crates


tomusdrw:

rphmeier:
This makes sense, but it would be nice to leave them in this repo as well, while linking them in with a default feature.

PR #76, opened 2017-11-09
Use ethereum-types library for uints and hashes

### LINKED ISSUE #69: Use `bigint` for uint types.

### LINKED ISSUE #84: How to convert U256 to stringified u64?
hey guys, probably a noob question. but i'd like to convert the `U256` gotten from the `balance` function to a human readable number. I assume a string u64 is the only way to do this?

i tried using the `low_u64` but i'm not even sure what that does, but the number it returns is wrong. 

code 👇 

```rust
extern crate web3;
use web3::*;
use types::{U256, Address};
use futures::{Future};

#[derive(Debug)]
struct Mapping {
	address: Address,
	balance: u64
}

impl Mapping {
	fn new (address: Address, balance: U256) -> Mapping {
		Mapping {
			address,
			balance: balance.low_u64()
		}
	}
}

fn main() {
	let (_eloop, transport) = web3::transports::Http::new("http://localhost:8545").unwrap();
	let web3 = web3::Web3::new(transport);
	let accounts = web3.eth().accounts().wait().unwrap();

	let balances: Vec<Mapping> = accounts.into_iter().fold(Vec::new(),|mut accum, address| {
		let balance = web3.eth().balance(address, None).wait();
		accum.push(Mapping::new(address, balance.unwrap()));
		accum
	});

	for balance in balances {
		println!("Balance : {:?}", balance)
	}
}
```

```bash
ganache-cli -e 30 -a 1
Ganache CLI v6.0.3 (ganache-core: 2.0.2)

Available Accounts
==================
(0) 0x6ce23c45eb5b818fdaf76944ba5dd5cf56990a83

Private Keys
==================
(0) 12059b88a7fd4db2e973e4c9670d344b119cd7f9b7f21324bf24245beb9eb4fc

HD Wallet
==================
Mnemonic:      image want animal chimney know dune age board world volcano great subway
Base HD Path:  m/44'/60'/0'/0/{account_index}

Listening on localhost:8545
eth_accounts
eth_getBalance

```

```bash
cargo run                                                                                                                                                                                    master ✚ ◼
   Compiling cryptobot v0.1.0 (file:///home/seunlanlege/Projects/cryptobot)
    Finished dev [unoptimized + debuginfo] target(s) in 2.97 secs
     Running `target/debug/cryptobot`
Balance : Mapping { address: 0x6ce23c45eb5b818fdaf76944ba5dd5cf56990a83, balance: 7766279631452241920 }



```

tomusdrw:
Closes #69 
Closes #84


debris:
isn't it better to go with `ethereum-types` instead?

tomusdrw:
Yes, I'd prefer to go with `uint` and `fixed-hash` libraries directly, although they are not yet on crates io.

hswick:
Hmmm would it be possible to FFI with a C/C++ library like this https://sourceforge.net/projects/ttmath/?

debris:
@hswick to be honest I see no point in using `tmath`. `rust-web3` library does little to none math operations and bringing C++ library as a dependency is just a bit of a hassle

hswick:
@debris I agree the C++ lib doesn't belong in this repo, however at some point the lack of crates problem needs to be solved. So wrapping an existing C++ seems like the path of least resistance.

forrest-marshall:
I'm hesitant about the idea of using `bigint` or its derivatives by default.  Given how foundational the `web3` crate is, I think it's goal should be to represent the jsonrpc APIs as transparently as possible (within reason).  Many consumers are either going to explicitly require canonical byte representations (e.g. hashers), or be indifferent to receiving a math enabled type (e.g. comparison ops).  The `bigint` crate is serializing and deserializing via intermediate arrays and vectors respectively anyhow.  I think ensuring that conversion to math enabled types is simple and easy would be more in line with rust's 'only pay for what your use' philosophy.

Issue #16 proposes breaking the transports out into a separate crate.  Perhaps the `types` module could become a reexport of a separate crate with minimal dependencies.  This would allow the authors of crates like `ethereum-types` to implement web3 specific conversions without adding unnecessary dependencies, while also allowing web3's consumers to maximize those sweet sweet zero-cost abstractions :)

debris:
>  I think ensuring that conversion to math enabled types is simple and easy would be more in line with rust's 'only pay for what your use' philosophy.

I agree, but this is a bit of a hassle for us to maintain. Not long ago, we had 4 or 5 different implementations of `uint` and `hash` being used in [parity](https://github.com/paritytech/parity). Converting between them was just super annoying. We've decided to 'sacrifice' modularity to simplify development.

> Issue #16 proposes breaking the transports out into a separate crate. Perhaps the types module could become a reexport of a separate crate with minimal dependencies. This would allow the authors of crates like ethereum-types to implement web3 specific conversions without adding unnecessary dependencies, while also allowing web3's consumers to maximize those sweet sweet zero-cost abstractions :)

`ethereum-types` and `web3` authors are the same people ;) And these 'ethereum' and 'web3' conversions are the same thing :)

forrest-marshall:
> I agree, but this is a bit of a hassle for us to maintain. Not long ago, we had 4 or 5 different implementations of uint and hash being used in parity. Converting between them was just super annoying. We've decided to 'sacrifice' modularity to simplify development.

Good point.  I can see how that would have serious diminishing returns.

> `ethereum-types` and `web3` authors are the same people ;) And these 'ethereum' and 'web3' conversions are the same thing :)

Haha, fair enough.  In defense of my sanity, the comment about unnecessary conversions was based on tracing minimum steps from hex to canonical byte repr (from point of view of caller) in current master branches:

- `ethereum-types`: hex buf -> [heap-allocated vector](https://github.com/paritytech/primitives/blob/c43a3bfed9484b8bed7ec5ea389bdc684015ef97/serialize/src/lib.rs#L97) (plus intermediate allocation of a formatting string if odd number of hex characters) -> [bigint internal representation](https://github.com/paritytech/primitives/blob/c43a3bfed9484b8bed7ec5ea389bdc684015ef97/uint/src/uint.rs#L945) -> [procedural write to slice](https://github.com/paritytech/primitives/blob/c43a3bfed9484b8bed7ec5ea389bdc684015ef97/uint/src/uint.rs#L696) (with indexing).

- `web3`: hex buf -> [stack-allocated array](https://github.com/tomusdrw/rust-web3/blob/f6f861285fde79aadac5fcb041daa379efd7b0b2/src/types/uint.rs#L161) (via iterator, so no bunds-checking).

tomusdrw:
My 2cents. The goal of `web3` is to provide simple to use ethereum integration, so most ethereum-specific operations should be possible without using any 3rd party libraries, as such I think that stuff like:
 - Incrementing the nonce
 - Calculating the gas price (from estimated gas price)
 - Calculating the gas
 should be possible and requires U256 arithmetics

Otherwise people may choose any jsonrpc-client (probably better than web3) since there is not that much we add on top of it.

Also `web3` is extensible, so it should be easy to create your own API that deserializes directly to `[u8; 32]` if you prefer. That should cover all cases where extreme performance is required.
I'm also open to simplify that even further (maybe exposing a method like: `let x: [u8; 32] = web3.raw_request("eth_getBalance", Address::from(5)).wait().unwrap()`)

Feel free to also contribute to `ethereum-types` to get rid of unnecessary allocations :)

tomusdrw:
@forrest-marshall https://github.com/paritytech/primitives/pull/15 ;)

forrest-marshall:
@tomusdrw well played :)

src/contract/tokens.rs
```diff
@@ -147,13 +147,13 @@ impl Tokenizable for H256 {
 impl Tokenizable for Address {
   fn from_token(token: Token) -> Result<Self, Error> {
     match token {
-      Token::Address(data) => Ok(types::H160(data)),
+      Token::Address(data) => Ok(data.into()),
```
debris: in `ethabi 5.1` `data` is already an `Address`, there is no need to call `.into()`, [docs](https://docs.rs/ethabi/5.1.0/ethabi/enum.Token.html#variant.Address)

Cargo.toml
```diff
@@ -11,17 +11,18 @@ authors = ["Tomasz Drwięga <tomasz@ethcore.io>"]
 
 [dependencies]
 arrayvec = "0.3"
+error-chain = "0.11.0"
 ethabi = "4.0"
```
debris: please bump `ethabi` to `5.1`

debris:
and let's merge it tomorrow! :)

debris:
there are still some conflicts ;)

PR #79, opened 2017-11-17
Example with Solidity

hswick:
Added full example with smart contract, example queries contract and changes state of contract, documented solidity compiling on README


examples/simple_storage.rs
```diff
@@ -0,0 +1,47 @@
...
+
+  //Get current balance
+  let balance = web3.eth().balance(accounts[0], None).wait().unwrap();
+
+  println!("Balance: {}", balance);
+
+  // Get the contract bytecode for instance from Solidity compiler
+  let bytecode: Vec<u8> = include_str!("/build/SimpleStorage.bin").from_hex().unwrap();
+  // Deploying a contract
+  let contract = Contract::deploy(web3.eth(), include_bytes!("/build/SimpleStorage.abi")).unwrap()
```
tomusdrw: Could you change the paths to be relative?

tomusdrw:
Looks good, thank you for the contribution. Could you update the paths to be relative? This should make CI happy :)

hswick:
@tomusdrw could you specify what you mean by relative? I thought they already were.

tomusdrw:
@hswick A dot is missing before first leading slash
```
include_bytes!("/build/SimpleStorage.abi")
// should be
include_bytes!("./build/SimpleStorage.abi")
```

You can see that tests fail because the path cannot be found (it's looking for relative path of `/build/SimplStorage.abi`

hswick:
Ah yes I see, wasn't sure if it was that obvious or not

tomusdrw:
Thanks!

PR #86, opened 2018-01-13
confirmation support for raw transactions

forrest-marshall:
Added support for confirmation of raw transactions.

The underlying confirmation mechanism for regular transactions works fine for raw transactions too.  Added a new constructor and public function so that this functionality is now accessible.


forrest-marshall:
Oops, missed that one.  Fixed.

PR #88, opened 2018-01-22
Workaround filter type.

### LINKED ISSUE #87: eth_getFilterChanges not working because of malformed id
Trying to capture events from a Smart Contract, I registered a Filter via `EthFilter.create_logs_filter(...)`.
The rpc responded with the filter id "0x03" which is internally transformed to a "3" as the id of the BaseFilter  struct. 
So when the filter changes are queried, it tries to query filter "0x3" instead of "0x**0**3", omitting the leading zero, which results in the rpc error: "FilterSubprovider - no filter with that id: 0x3".
It is possible to fix this by changing the BaseFilter id to type "String", but that might not be wanted.

This is the according output from ganache-cli:

```
# The library requests a new Filter:
eth_newFilter
   > {
   >   "jsonrpc": "2.0",
   >   "method": "eth_newFilter",
   >   "params": [
   >     {
   >       ...
   >     }
   >   ],
   >   "id": 1,
   >   "external": true
   > }
...
# Response containing the Filter ID 0x03:
 <   {
 <     "id": 1,
 <     "jsonrpc": "2.0",
 <     "result": "0x03"
 <   }
...
# Filter Changes fail because of wrong ID, always returning empty JSON Array in the end
eth_getFilterChanges
   > {
   >   "jsonrpc": "2.0",
   >   "method": "eth_getFilterChanges",
   >   "params": [
   >     "0x3"
   >   ],
   >   "id": 4,
   >   "external": true
   > }
FilterSubprovider - no filter with that id: 0x3
 <   {
 <     "id": 4,
 <     "jsonrpc": "2.0",
 <     "result": []
 <   }

```

Best regards,
hardliner93

tomusdrw:
Closes #87

### DIFF
--- a/examples/simple_storage.rs
+++ b/examples/simple_storage.rs
@@ -4,7 +4,7 @@ extern crate rustc_hex;
 
 use web3::futures::Future;
 use web3::contract::{Contract, Options};
-use web3::types::{Address, U256};
+use web3::types::U256;
 use rustc_hex::FromHex;
 use std::time;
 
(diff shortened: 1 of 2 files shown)


tomusdrw:
Should be fixed now, I just lazily ran tests on the CI :)

PR #91, opened 2018-01-30
Add proper return type for `eth_syncing`.

forrest-marshall:
Added the expected return type for `eth_syncing` as per the ethereum jsonrpc spec.  Tested against syncing and non-syncing parity node.  This is a breaking api change, but I believe this (or something like it) is necessary pre 1.0, since the current return value (bool) is definitely insufficient.  Cheers.

*edit*:  Forgot that I horribly mangled the tests.  Fixed.

### DIFF
--- a/src/api/eth.rs
+++ b/src/api/eth.rs
@@ -6,7 +6,7 @@ use types::{
   Address, Block, BlockId, BlockNumber, Bytes, CallRequest,
   H64, H256, H520, Index,
   Transaction, TransactionId, TransactionReceipt, TransactionRequest,
-  U256, Work,
+  U256, Work, SyncState,
 };
 use {Transport};
 
@@ -301,9 +301,8 @@ impl<T: Transport> Eth<T> {
     CallResult::new(self.transport.execute("eth_submitWork", vec![nonce, pow_hash, mix_hash]))
   }
 
-  // TODO [ToDr] Proper type?
   /// Get syncing status
-  pub fn syncing(&self) -> CallResult<bool, T::Out> {
+  pub fn syncing(&self) -> CallResult<SyncState, T::Out> {
     CallResult::new(self.transport.execute("eth_syncing", vec![]))
   }
 }
--- a/src/api/eth.rs
+++ b/src/api/eth.rs
@@ -670,7 +670,12 @@ mod tests {
   );
 
   rpc_test! (
-    Eth:syncing => "eth_syncing";
+    Eth:syncing:syncing => "eth_syncing";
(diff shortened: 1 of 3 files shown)


src/types/sync_state.rs
```diff
@@ -0,0 +1,74 @@
...
+
+/// implementation detail for sync state deserialization
+#[derive(Debug,Serialize,Deserialize)]
+#[serde(untagged)]
+enum Either<A,B> {
+    A(A),
+    B(B)
+}
+
+
```
tomusdrw: Bunch of superflous lines.

src/types/sync_state.rs
```diff
@@ -0,0 +1,74 @@
...
+// The `eth_syncing` method returns either `false` or an instance of the sync info object.
+// This doesn't play particularly well with the features exposed by `serde_derive`,
+// so we use the custom impls below to ensure proper behavior.
+
+impl<'de> Deserialize<'de> for SyncState {
+
+    fn deserialize<D>(deserializer: D) -> Result<Self,D::Error> where D: Deserializer<'de> {
+        let either: Either<SyncInfo,bool> = Deserialize::deserialize(deserializer)?;
```
tomusdrw: Neat solution!

src/types/sync_state.rs
```diff
@@ -0,0 +1,74 @@
+use serde::de::{Deserialize,Deserializer,Error};
+use serde::ser::{Serialize,Serializer};
+use types::U256;
+
+/// Information about current blockchain syncing operations.
+#[derive(Debug,Clone,PartialEq,Serialize,Deserialize)]
```
tomusdrw: `s/,/, /g`

src/api/eth.rs
```diff
@@ -672,6 +671,6 @@ mod tests {
 
   rpc_test! (
     Eth:syncing => "eth_syncing";
-    Value::Bool(true) => true
+    json!({"startingBlock": "0x384","currentBlock": "0x386","highestBlock": "0x454"}) => SyncState::Syncing(SyncInfo { starting_block: 0x384.into(), current_block: 0x386.into(), highest_block: 0x454.into()})
```
tomusdrw: Could we have a test for `false` as well?

tomusdrw:
Thanks!

forrest-marshall:
Happy to help.  Using this crate fairly extensively in my codebase.  Expect a few more PRs for potential fixes/improvements over the next few weeks.  Cheers.

src/types/sync_state.rs
forrest-marshall: Thanks :)

PR #93, opened 2018-02-01
fix peer count return type

forrest-marshall:
The `Net::peer_count` method was returning a string.  Changed to `U256` to reflect expected return type as per spec.  Updated unit tests in crate and ran sanity checks against local parity node.

@tomusdrw I'll probably have a few more of these over the next couple weeks.  In the process of integrating `web3` into a number of components of a mid size codebase.  Let me know if you would prefer reviewing a batch PR.  Otherwise I'll be keeping it per-feature.  Cheers.

### DIFF
--- a/src/api/net.rs
+++ b/src/api/net.rs
@@ -2,6 +2,7 @@
 
 use api::Namespace;
 use helpers::CallResult;
+use types::U256;
 
 use {Transport};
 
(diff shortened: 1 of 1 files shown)


tomusdrw:
Thanks! Small PRs are better for me to review. Seems that the return type in Parity is incorrect (it's `String`, but returns a serialized number 0.o) and that API was based on this.

Issue #97, opened 2018-02-15
Receipt should include status field


akuanti:
The current version of JSON RPC includes the following for [`eth_getTransactionReceipt`](https://github.com/ethereum/wiki/wiki/JSON-RPC#eth_gettransactionreceipt)

> It also returns either :
>
> `root` :` DATA` 32 bytes of post-transaction stateroot (pre Byzantium)
> `status`: `QUANTITY` either 1 (success) or 0 (failure)

The `Receipt` struct does not include either of these fields.

tomusdrw:
Resolved via #98

PR #98, opened 2018-02-16
Added status field to Receipt.

akuanti:
Addresses #97. I'm not sure if this is the correct way to deal with the possibility of either `status` or `root`.

### DIFF
--- a/src/types/transaction.rs
+++ b/src/types/transaction.rs
@@ -57,6 +57,8 @@ pub struct Receipt {
   pub contract_address: Option<H160>,
   /// Logs generated within this transaction.
   pub logs: Vec<Log>,
+  /// Status: either 1 (success) or 0 (failure).
+  pub status: Option<u32>,
 }
 
 #[cfg(test)]
--- a/src/types/transaction.rs
+++ b/src/types/transaction.rs
@@ -1,4 +1,4 @@
-use types::{H160, H256, U256, Index, Log, Bytes};
+use types::{H160, H256, U128, U256, Index, Log, Bytes};
 
 /// Description of a Transaction, pending or in the chain.
 #[derive(Debug, Default, Clone, PartialEq, Deserialize)]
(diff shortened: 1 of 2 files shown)


src/types/transaction.rs
```diff
@@ -57,6 +57,8 @@ pub struct Receipt {
   pub contract_address: Option<H160>,
   /// Logs generated within this transaction.
   pub logs: Vec<Log>,
+  /// Status: either 1 (success) or 0 (failure).
+  pub status: Option<u32>,
```
tomusdrw: Since the spec says `QUANTITY` it should be encoded as hex. To achieve that we need to use a type from `ethereum_types`. I think the smallest there is `U64` so we should use this.

Corresponding code from Parity:
https://github.com/paritytech/parity/blob/afae75b57cd70d81e968fc66319431871a79ff3d/rpc/src/v1/types/receipt.rs#L53

akuanti:
Do you mean `U64` or `H64`? The smallest `uint` I see in `ethereum-types` is `U128`, and as far as I can tell, Parity is using its own internal `U64` type. If I use `H64`, I get an error about the length of the string being only 1.

This actually brings up something I've been wondering, which is, what is the relationship between this project and Parity? Is seems like there is quite a bit of overlap between them, particularly the RPC types, but they aren't in sync. Does it make sense to extract these types into a separate crate, kind of like `ethereum-types`? That way, changes to the RPC spec propagate to both projects.

tomusdrw:
https://github.com/paritytech/primitives/blob/master/ethereum-types/src/uint.rs#L7 There is U64 in ethereum types, perhaps we didn't release a new version yet.

Parity and Web3 are both written in Rust and both communicate with each other. Sharing the types indeed makes perfect sense, although the licenses are currently not compatible. Extracting `ethereum-types` like library would solve that.

akuanti:
Okay, I'll update the type when the new version is released.

> Sharing the types indeed makes perfect sense, although the licenses are currently not compatible. Extracting ethereum-types like library would solve that.

Is there anything I can do to help in this effort?

Issue #99, opened 2018-02-21
Log.log_type should be Log.removed


Kmoneal:
According to the JSON RPC, type: String is no longer used and logs now return with removed: bool. https://github.com/ethereum/wiki/wiki/JSON-RPC#returns-42

tomusdrw:
Closed via #100

PR #100, opened 2018-02-27
Updated Log Struct

Kmoneal:
Addresses [#99](https://github.com/tomusdrw/rust-web3/issues/99).


Kmoneal:
Any preference on naming?

tomusdrw:
What I mean is something like that:
```
struct Log {
  ...
  /// type
  #[serde(rename="logType")]
  pub log_type: Option<String>,
  /// Removed
  pub removed: Option<bool>,
```

So both parameters are supported, but marked as optional. An additional `fn is_removed(&self) -> bool` helper method can be provided to cover both cases.

Issue #37, opened 2017-05-30
Pub-Sub support


tomusdrw:
Requires #14

rphmeier:
```rust
trait DuplexTransport: Transport {
    type Stream: Stream<Item=rpc::Value, Error=Error>; // drop should unsubscribe.
    type Future: Future<Item=Self::Stream, Error=Error>;

    fn subscribe(request) -> Self::Future;
}

// example method implementation
let stream = self.transport.subscribe(my_subscription_request).and_then(|s| s.map(lazy_decode));
```

How should it work with batch transport? Should you be able to batch regular requests with subscriptions?

Maybe a general stream wrapper for encapsulating subscription ID and dropping logic?

tomusdrw:
Resolved via #105

PR #101, opened 2018-03-06
Add WebSocket transport

mjkoo:
Addresses #14 , gated behind "ws" feature (did add to default features however), adds https://crates.io/crates/websocket as a dependency.

Tested basic RPC functionality against geth 1.8.2 with the websocket endpoint enabled.

### DIFF
--- a/src/transports/ws.rs
+++ b/src/transports/ws.rs
@@ -64,7 +64,6 @@ impl WebSocket {
       let write_sender_ = write_sender.clone();
 
       ClientBuilder::from_url(&url)
-        .add_protocol("web3.rs")
         .async_connect(None, handle)
         .from_err::<Error>()
         .map(|(duplex, _)| duplex.split())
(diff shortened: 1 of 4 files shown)


src/transports/ws.rs
```diff
@@ -0,0 +1,258 @@
...
+    let url: Url = url.parse()?;
+    let (write_sender, write_receiver) = mpsc::unbounded();
+    let pending: Arc<Mutex<BTreeMap<RequestId, Pending>>> = Arc::new(Mutex::new(BTreeMap::new()));
+
+    let ws_future = {
+      let pending_ = pending.clone();
+      let write_sender_ = write_sender.clone();
+
+      ClientBuilder::from_url(&url)
+        .add_protocol("web3.rs")
```
tomusdrw: Is the protocol required for anything in other clients? Parity will reject such requests, it expects protocol to be empty. Can we remove that?

mjkoo:
Looks good to me, both changes should be taken care of in that last commit :)

If this looks good to you I'd like to take a look at #37 next

tomusdrw:
@mjkoo perfect! Let me know if you need any help with #37

PR #102, opened 2018-03-07
Added contract::call_with_confirmations().

akuanti:
Allows you to call a contract function and wait for a specified number of confirmations.

### DIFF
--- a/src/contract/mod.rs
+++ b/src/contract/mod.rs
@@ -3,7 +3,8 @@
 use ethabi;
 
 use std::time;
-use api::Eth;
+use api::{Eth, Namespace};
+use confirm;
 use contract::tokens::{Detokenize, Tokenize};
 use types::{Address, Bytes, CallRequest, H256, TransactionRequest, TransactionCondition, U256, BlockNumber};
 use {Transport};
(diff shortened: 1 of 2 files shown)


src/contract/mod.rs
```diff
@@ -106,6 +107,31 @@ impl<T: Transport> Contract<T> {
...
+  /// Execute a contract function and wait for confirmations
+  pub fn call_with_confirmations<P>(&self, func: &str, params: P, from: Address, options: Options, confirmations: usize)
+    -> confirm::SendTransactionWithConfirmation<T>
+    where P: Tokenize
+  {
+    let poll_interval = time::Duration::from_secs(1);
+
+    let fn_data = self.abi.function(func.into())
+        .and_then(|function| function.encode_input(&params.into_tokens()))
+        .unwrap();
```
tomusdrw: Would be good to handle the error here.

akuanti: I'm not sure what is the right way to handle this (still getting used to Rust). Should the function return a `Result`?

tomusdrw: Have a look at: https://github.com/tomusdrw/rust-web3/blob/a9611d3f0fba51c6f40f51f8993c585e2a8a7ae5/src/contract/result.rs#L55-L63

`CallResult` in that case encapsulate an asynchronous result (positive `Ok()` and negative `Err()`).

We should probably have something similar for `SendTransactionWithConfirmation` type.
It will require introducing new state here:
https://github.com/tomusdrw/rust-web3/blob/a9611d3f0fba51c6f40f51f8993c585e2a8a7ae5/src/confirm.rs#L197
that immediately resolves to an error.

tomusdrw: Let me know if you need more help with that.

akuanti: Yeah, I think I do need more help with this. Are you saying that `call_with_confirmations()` should still return `SendTransactionWithConfirmation`, but that `SendTransactionWithConfirmation` should have an error state so that any errors that happen in the function can be converted into it?

tomusdrw: @akuanti I've made a PR to your branch to patch that. Although I'm not super happy with the solution,cause the error types does not match. We can cover that as part of #103 though.

PR #105, opened 2018-03-14
Pubsub

mjkoo:
Adds pubsub support to the websocket and ipc transports, addresses #37 . Made a couple decisions I'd like some feedback on in review, will annotate diff.

### DIFF
--- a/examples/pubsub.rs
+++ b/examples/pubsub.rs
@@ -1,0 +1,24 @@
+extern crate web3;
+
+use web3::futures::{Future, Stream};
+
+fn main() {
+    let (_eloop, ws) = web3::transports::WebSocket::new("ws://localhost:8546").unwrap();
+    let web3 = web3::Web3::new(ws.clone());
+    let mut sub = web3.eth_subscribe().subscribe_new_heads().wait().unwrap();
+
+    println!("Got subscription id: {:?}", &sub);
+
+    (&mut sub)
+        .take(5)
+        .for_each(|x| {
+            println!("Got: {:?}", x);
+            Ok(())
+        })
+        .wait()
+        .unwrap();
+
+    sub.unsubscribe();
+
+    drop(web3);
+}
--- a/examples/pubsub.rs
+++ b/examples/pubsub.rs
@@ -7,7 +7,7 @@ fn main() {
     let web3 = web3::Web3::new(ws.clone());
     let mut sub = web3.eth_subscribe().subscribe_new_heads().wait().unwrap();
 
-    println!("Got subscription id: {:?}", &sub);
+    println!("Got subscription id: {:?}", sub.id());
 
     (&mut sub)
         .take(5)
--- a/src/api/eth_subscribe.rs
+++ b/src/api/eth_subscribe.rs
@@ -1,0 +1,161 @@
+//! `Eth` namespace, subscriptions
+
+use std::fmt;
+use std::marker::PhantomData;
+
+use api::Namespace;
+use futures::{Async, Future, Poll, Stream};
+use helpers::{self, CallResult};
+use rpc;
+use serde;
+use serde_json;
+use types::{Address, BlockHeader, Filter, H256, Log};
+use {DuplexTransport, Error};
+
+/// `Eth` namespace, subscriptions
+#[derive(Debug, Clone)]
+pub struct EthSubscribe<T> {
+  transport: T,
+}
+
+impl<T: DuplexTransport> Namespace<T> for EthSubscribe<T> {
+  fn new(transport: T) -> Self
+  where
+    Self: Sized,
+  {
+    EthSubscribe { transport }
+  }
+
+  fn transport(&self) -> &T {
+    &self.transport
+  }
+}
+
+pub struct SubscriptionStream<T: DuplexTransport, I> {
+  transport: T,
+  id: String,
+  rx: T::NotificationStream,
+  _marker: PhantomData<I>,
+}
+
+impl<T: DuplexTransport, I> SubscriptionStream<T, I> {
+  pub fn new(transport: T, id: String) -> Self {
+    let rx = transport.subscribe(&id);
+    SubscriptionStream {
+      transport,
+      id,
+      rx,
+      _marker: PhantomData,
+    }
+  }
+
+  pub fn id(&self) -> &str {
+    &self.id
+  }
+}
+
+impl<T, I> Stream for SubscriptionStream<T, I>
+where
+  T: DuplexTransport,
+  I: serde::de::DeserializeOwned,
+{
+  type Item = I;
+  type Error = Error;
+
+  fn poll(&mut self) -> Poll<Option<Self::Item>, Self::Error> {
+    match self.rx.poll() {
+      Ok(Async::Ready(Some(x))) => serde_json::from_value(x)
+        .map(Async::Ready)
+        .map_err(Into::into),
+      Ok(Async::Ready(None)) => Ok(Async::Ready(None)),
+      Ok(Async::NotReady) => Ok(Async::NotReady),
+      Err(e) => Err(e),
+    }
+  }
+}
+
+impl<T: DuplexTransport, I> fmt::Debug for SubscriptionStream<T, I> {
+  fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+    write!(f, "{:?}", self.id())
+  }
+}
+
+impl<T: DuplexTransport, I> Drop for SubscriptionStream<T, I> {
+  fn drop(&mut self) {
+    self.transport.unsubscribe(self.id());
+  }
+}
+
+#[derive(Debug)]
+pub struct SubscriptionResult<T: DuplexTransport, I> {
+  transport: T,
+  inner: CallResult<String, T::Out>,
+  _marker: PhantomData<I>,
+}
+
+impl<T: DuplexTransport, I> SubscriptionResult<T, I> {
+  pub fn new(transport: T, id_future: CallResult<String, T::Out>) -> Self {
+    SubscriptionResult {
+      transport,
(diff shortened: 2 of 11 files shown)


src/api/eth_subscribe.rs
```diff
@@ -0,0 +1,161 @@
...
+
+impl<T: DuplexTransport, I> fmt::Debug for SubscriptionStream<T, I> {
+  fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+    write!(f, "{:?}", self.id())
+  }
+}
+
+impl<T: DuplexTransport, I> Drop for SubscriptionStream<T, I> {
+  fn drop(&mut self) {
+    self.transport.unsubscribe(self.id());
```
mjkoo: In ticket for #37 , desired behavior was to unsubscribe from notifications on dropping the stream. This "unregisters" this stream from the transport, but does not invoke `eth_unsubscribe` via rpc. Unsure how to best handle this, thought waiting on a future from drop would be a bad idea. Currently if a user forgets to call eth_unsubscribe it's not fatal but notifications will continue to be sent via the transport and not forwarded to a stream.

src/types/block.rs
```diff
@@ -1,6 +1,49 @@
 use serde::{Serialize, Serializer};
 use types::{Bytes, U128, U256, H256, H160, H2048};
 
+/// The block type returned from RPC calls.
+#[derive(Debug, Clone, PartialEq, Deserialize)]
+pub struct BlockHeader {
```
mjkoo: Duplication here with the Block struct, added this "header" struct to avoid making breaking changes, possible alternatives would be to make the remaining fields `Option<...>`s or make the Block struct contain a BlockHeader which likely would need a custom Deserialize implementation and also would change the struct layout.

src/api/eth_subscribe.rs
```diff
@@ -0,0 +1,161 @@
...
+  }
+
+  fn transport(&self) -> &T {
+    &self.transport
+  }
+}
+
+pub struct SubscriptionStream<T: DuplexTransport, I> {
+  transport: T,
+  id: String,
```
tomusdrw: Can we extract `Id` to a typedef? Will be easier to read than a bunch of `String`s

src/api/eth_subscribe.rs
```diff
@@ -0,0 +1,161 @@
...
+  {
+    EthSubscribe { transport }
+  }
+
+  fn transport(&self) -> &T {
+    &self.transport
+  }
+}
+
+pub struct SubscriptionStream<T: DuplexTransport, I> {
```
tomusdrw: Missing docs for public types

src/api/eth_subscribe.rs
tomusdrw: Can we get a `pub fn unsubscribe(self)` on the stream to allow people to unsubscribe?

src/api/eth_subscribe.rs
```diff
@@ -0,0 +1,161 @@
...
+    }
+  }
+}
+
+impl<T, I> Future for SubscriptionResult<T, I>
+where
+  T: DuplexTransport,
+  I: serde::de::DeserializeOwned,
+{
+  type Item = (String, SubscriptionStream<T, I>);
```
tomusdrw: Do we really need the `id` as a result?

src/api/eth_subscribe.rs
```diff
@@ -0,0 +1,161 @@
...
+    SubscriptionResult::new(self.transport().clone(), id_future)
+  }
+
+  pub fn subscribe_syncing(&self) -> SubscriptionResult<T, rpc::Value> {
+    let subscription = helpers::serialize(&&"syncing");
+    let id_future = CallResult::new(self.transport.execute("eth_subscribe", vec![subscription]));
+    SubscriptionResult::new(self.transport().clone(), id_future)
+  }
+
+  pub fn unsubscribe(&self, id: &str) -> CallResult<bool, T::Out> {
```
tomusdrw: I would rather see people using `Stream::unsubscribe()` instead, hence imho better to hide the id

src/api/eth_subscribe.rs
tomusdrw: Actually maybe best to have a wrapper type:
```
#[derive(...)]
pub struct Id(String)
```
to make all interactions with id typesafe.

src/transports/ipc.rs
```diff
@@ -155,6 +161,20 @@ impl BatchTransport for Ipc {
   }
 }
 
+impl DuplexTransport for Ipc {
+  type NotificationStream = Box<Stream<Item = rpc::Value, Error = Error> + Send + 'static>;
+
+  fn subscribe(&self, id: &str) -> Self::NotificationStream {
+    let (tx, rx) = mpsc::unbounded();
+    self.subscriptions.lock().insert(id.to_owned(), tx);
```
tomusdrw: Should we care about ID conflict here? Panic/Error?

src/transports/ipc.rs
```diff
@@ -260,40 +281,69 @@ impl Future for ReadStream {
...
+
+          if let (Some(&rpc::Value::String(ref id)), Some(result)) = (id, result) {
+            if let Some(stream) = self.subscriptions.lock().get(id) {
+              if let Err(e) = stream.unbounded_send(result.clone()) {
+                error!("Error sending notification (id: {:?}): {:?}", id, e);
+              }
+            } else {
+              warn!("Got notification for unknown subscription (id: {:?})", id);
+            }
+          }
```
tomusdrw: Some additional warnings here?

src/transports/ws.rs
```diff
@@ -56,62 +59,82 @@ impl WebSocket {
     trace!("Connecting to: {:?}", url);
 
     let url: Url = url.parse()?;
-    let (write_sender, write_receiver) = mpsc::unbounded();
     let pending: Arc<Mutex<BTreeMap<RequestId, Pending>>> = Arc::new(Mutex::new(BTreeMap::new()));
+    let subscriptions: Arc<Mutex<BTreeMap<String, Subscription>>> =
```
tomusdrw: `let subscriptions: <type> = Default::default();` ?

src/transports/ws.rs
```diff
@@ -189,6 +213,20 @@ impl BatchTransport for WebSocket {
   }
 }
 
+impl DuplexTransport for WebSocket {
+  type NotificationStream = Box<Stream<Item = rpc::Value, Error = Error> + Send + 'static>;
+
+  fn subscribe(&self, id: &str) -> Self::NotificationStream {
+    let (tx, rx) = mpsc::unbounded();
+    self.subscriptions.lock().insert(id.to_owned(), tx);
```
tomusdrw: Same as in IPC transport, what about conflicts?

tomusdrw:
One more thing: Can we get an example of using subscriptions?

examples/pubsub.rs
```diff
@@ -0,0 +1,27 @@
...
+fn main() {
+    env_logger::init();
+
+    let (_eloop, ws) = web3::transports::WebSocket::new("ws://localhost:8546").unwrap();
+    let web3 = web3::Web3::new(ws.clone());
+    let mut sub = web3.eth_subscribe().subscribe_new_heads().wait().unwrap();
+
+    println!("Got subscription id: {:?}", &sub);
+
+    (&mut sub)
```
mjkoo: Part of the reason for exposing subscription IDs in the initial PR was because this type of interface didn't seem very ergonomic to me but seemed to be needed to make the borrow checker happy. If there is a better way to do this let me know.

I think this won't be a problem with the upcoming futures 0.2 as for_each returns a future which resolves to the original stream, allowing you to call methods on it to clean stuff up, in the current version it resolves to unit.

mjkoo:
Rebased on current master and addressed issues in review

src/api/eth_subscribe.rs
```diff
@@ -0,0 +1,174 @@
...
+    &self,
+    filter: Filter
+  ) -> SubscriptionResult<T, Log> {
+    let subscription = helpers::serialize(&&"logs");
+    let filter = helpers::serialize(&filter);
+    let id_future = CallResult::new(self.transport.execute("eth_subscribe", vec![subscription, filter]));
+    SubscriptionResult::new(self.transport().clone(), id_future)
+  }
+
+  pub fn subscribe_new_pending_transactions(&self) -> SubscriptionResult<T, H256> {
```
tomusdrw: Missing docs

src/api/eth_subscribe.rs
```diff
@@ -0,0 +1,174 @@
...
+impl From<String> for SubscriptionId {
+  fn from(s: String) -> Self {
+    SubscriptionId(s)
+  }
+}
+
+/// Stream of notifications from a subscription
+/// Given a type deserializable from rpc::Value and a subscription id, yields items of that type as
+/// notifications are delivered.
+pub struct SubscriptionStream<T: DuplexTransport, I> {
```
tomusdrw: Can we `#[derive(Debug)]` here?

src/api/eth_subscribe.rs
```diff
@@ -0,0 +1,174 @@
...
+    }
+  }
+
+  /// Return the ID of this subscription
+  pub fn id(&self) -> &SubscriptionId {
+    &self.id
+  }
+
+  /// Unsubscribe from the event represented by this stream
+  pub fn unsubscribe(&self) -> CallResult<bool, T::Out> {
```
tomusdrw: Is it possible to consume `self` here instead of allowing to call multiple times?

tomusdrw:
Thank you so much @mjkoo! Great work!

Just waiting for an update of `ethereum-types` and `ethabi` and will be releasing new version of web3 soon!

mjkoo:
Awesome, thanks!

Issue #90, opened 2018-01-27
Doesn't support TLS


richardartoul:
The HTTP transport does not support TLS:

```rust
let (_eloop, transport) = web3::transports::Http::new("https://sokol.poa.network").unwrap();
```

leads to

```bash
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Transport("Io(Error { repr: Custom(Custom { kind: InvalidInput, error: NotHttp }) })")', src/libcore/result.rs:906:4
note: Run with `RUST_BACKTRACE=1` for a backtrace.
```

tomusdrw:
Right, we are using `hyper` directly. We should probably integrate https://docs.rs/hyper-tls/0.1.2/hyper_tls/ under a feature flag.

SeunLanLege:
would like to help out, i'm very familiar with the hyper crate

tomusdrw:
Resolved via #106

PR #106, opened 2018-03-15
Add TLS support to the HTTP transport

mjkoo:
Addresses #90 , tested against infura's https endpoint, also against geth http with the tls flag on and off.


mjkoo:
Both IPC tests are failing on travis but passing locally, seems like WouldBlock when reading from the unix socket isn't handled. This PR doesn't touch the IPC transport but still am observing failures, there is a fix for both tests in https://github.com/tomusdrw/rust-web3/pull/105/commits/aca43dfd8f07306b882436a7ca5474eb59f472c5

PR #112, opened 2018-04-07
Use auth credentials if specified in URL

blackbeam:
Because basic auth credentials are [ignored by hyper](https://github.com/hyperium/hyper/issues/854).


tomusdrw:
Thanks!

Issue #117, opened 2018-05-01
Log subscription topic filters not working


mjkoo:
`eth_subscribe` takes parameters which appear similar to a `web3::types::Filter`, but aren't, their format is described at https://github.com/ethereum/go-ethereum/wiki/RPC-PUB-SUB#logs . Passing filters in the same format as `eth_getLogs` et. al. results in no events being delivered.

tomusdrw:
Fixed via #118

PR #118, opened 2018-05-01
Fix subscribe_logs to pass topics correctly

mjkoo:
Addresses #117 , reference https://github.com/ethereum/go-ethereum/wiki/RPC-PUB-SUB#logs for parameter types


mjkoo:
Still seeing some strange behavior from geth in testing with this, am looking into it more. Any insight into how to correctly `eth_subscribe` to logs?

src/api/eth_subscribe.rs
```diff
@@ -147,12 +147,32 @@ impl<T: DuplexTransport> EthSubscribe<T> {
...
+        enum AddressVariants {
+            None,
+            Address(Address),
+            Addresses(Vec<Address>),
+        }
+
+        #[derive(Serialize)]
+        struct SubscribeLogParameters {
+            address: AddressVariants,
+            topics: Option<Vec<H256>>,
```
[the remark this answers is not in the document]
mjkoo: I'm doing some tests on it now, will have an updated PR in a couple mins. Seems like there may be larger geth compatibility issues with filters

src/types/log.rs
```diff
@@ -91,13 +114,21 @@ impl FilterBuilder {
...
     pub fn address(mut self, address: Vec<H160>) -> Self {
-        self.filter.address = Some(address);
+        self.filter.address = Some(ValueOrArray(address));
         self
     }
 
     /// Topics
     pub fn topics(mut self, topic1: Option<Vec<H256>>, topic2: Option<Vec<H256>>, topic3: Option<Vec<H256>>, topic4: Option<Vec<H256>>) -> Self {
-        self.filter.topics = Some(vec![topic1, topic2, topic3, topic4]);
+        let mut topics = vec![topic1, topic2, topic3, topic4]
```
mjkoo: Somewhat of a compromise to reduce breakage in the interface here, trims off the trailing `None` values from the topics as a shorter topic array will match regardless.

Think ideally this would be e.g. `pub fn topics(mut self, topcs: Vec<Option<Vec<H256>>>) -> Self` with perhaps `build()` checking that `topics.len() <= 4`, thoughts?

mjkoo:
Did some experimentation with geth, added two new examples showing listening for event logs via `EthFilter::create_logs_filter` and `EthSubscribe::subscribe_logs`. These did not work  with geth when linked with the released crate, but do with these changes.

Determined the following in serialized filters were causing issues with geth:
- `[TOPIC_HASH, null, null, null]` won't match logs that are `[TOPIC_HASH]`, that is null values at the end of the topics array prevent geth from matching
- `[[TOPIC_HASH], ...]` won't match, geth wants single value topics to not be nested in arrays (also true for addresses)
- Filter criteria that aren't present but serialized as `null` (e.g. fromBlock et. al.) will prevent matching

Restored `subscribe_logs` to take a filter, changed filter serialization to respect these rules. Used two examples added as tests against geth. Did not test against parity etc.

Question about change to FilterBuilder api added as comment in review.

src/types/log.rs
tomusdrw: Makes sense as well. I hope that Rust will allow us to omit `None` soon in the future by introducing optional parameters, though :)

PR #122, opened 2018-05-18
Add support for eth_getLogs method.

forrest-marshall:
Added the `eth_getLogs` method to the `Eth` namespace.  Tested against parity.

Side note: Is there a reason why all the methods take ownership of their arguments, even though they just pass references to `helpers::serialize`?  I conformed to the style but it seems wasteful, especially for methods like `Eth::sign`, which unnecessarily consume heap-allocated buffers.  It may be worth transitioning to something along the lines of `T: Borrow<Foo>` in some future version (should be a non-breaking change for most callers).

### DIFF
--- a/src/api/eth.rs
+++ b/src/api/eth.rs
@@ -2,7 +2,7 @@
 
 use api::Namespace;
 use helpers::{self, CallResult};
-use types::{Address, Block, BlockId, BlockNumber, Bytes, CallRequest, H256, H520, H64, Index, SyncState, Transaction, TransactionId, TransactionReceipt, TransactionRequest, U256, Work};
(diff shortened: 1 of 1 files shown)


forrest-marshall:
Hmm, am I missing something here?  Looks like CI build failed on previously passing format checks unrelated to this PR.  Changes in `rustfmt` behavior since last build maybe?

snd:
that seems to be the issue. i just updated stable locally and got the same result.

@tomusdrw we should probably pin rustfmt or even stable somehow on travis so breakages like this don't happen due to changes outside our control. i propose we do that and fix the formatting errors in a separate PR, then @forrest-marshall can rebase on master once that's merged and we can get this PR in. what do you think? let me know and i'll do it

Issue #81, opened 2017-12-12
Not compiling on Windows, 1.24 nightly


regresscheck:
Hi.
I'm getting this when trying to build example:
```
error[E0432]: unresolved import `self::tokio_uds::UnixStream`
  --> C:\Users\Andrey\.cargo\git\checkouts\rust-web3-ebc1a09fd5a9a2a2\c6bd1ab\src\transports\ipc.rs:10:5
   |
10 | use self::tokio_uds::UnixStream;
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^ no `UnixStream` in the root

error: aborting due to previous error

error: Could not compile `web3`.

```

Tried both gnu and msvc toolchains, cargo and github versions. Am I missing something?

tomusdrw:
IPC feature is enabled by default, but is not available on windows. Could you try reference the library like this:
```
web3 = { git = "https://github.com/tomusdrw/rust-web3.git", default-features = false, features = ["http"] }
```

regresscheck:
Thanks! Everything is fine now. Can you please mention it somewhere, so I would be the only one creating such issue? :)

tomusdrw:
Answers are now in this thread:) I would really appreciate a PR to the readme file mentioning windows build though.

afck:
This way, every dependent crate has to do a lot of conditional compilation and have a complicated `Cargo.toml` if it wants to offer IPC support on Unix but also run on Windows. Wouldn't it make more sense if `web3` itself compiled on Windows, but `Ipc::new` just always failed with an error message explaining that this is Unix-only? That would just involve a little bit more conditional compilation in `transports/mod.rs` and an `ipc_windows` dummy module that would do nothing expect returning that error.

If you are okay with this, please reopen this Issue, then I'll try and write the code.

Alternatively, the `ipc` module could just be omitted on Windows (but without failing to compile). That would still be less complicated for the user than it is now, because at least the `Cargo.toml` could just keep the `ipc` feature unconditionally.

tomusdrw:
@afck both solutions make sense to me, feel free to submit a PR

tomusdrw:
Fixed in #124

PR #124, opened 2018-05-22
Fix Windows build: Make Ipc::new return Err. (#81)

afck:
* The crate builds successfully on non-Unix systems now, even with
  the default features, including `ipc`. If IPC is not available,
  the constructor returns an error.

### DIFF
--- a/src/transports/ipc.rs
+++ b/src/transports/ipc.rs
@@ -1,5 +1,6 @@
 //! IPC Transport for *nix
 
+#[cfg(unix)]
 extern crate tokio_uds;
 
 use std::collections::BTreeMap;
@@ -7,6 +8,7 @@ use std::io::{self, Read, Write};
 use std::path::Path;
 use std::sync::{atomic, Arc};
 
+#[cfg(unix)]
 use self::tokio_uds::UnixStream;
 
 use api::SubscriptionId;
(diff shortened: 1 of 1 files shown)


afck:
@tomusdrw: I didn't think this through: I don't have access to a Windows machine and don't have a way to even test whether this compiles. Do you know how we can try it out? I wonder whether it would make sense to use AppVeyor in addition to Travis?

tomusdrw:
@afck Right, if we want to maintain windows compatibility long-term using appveyor to run builds would be good. Will try to have a look at this tonight, feel free to add appveyor configuration earlier.

afck:
Looks like we have another problem: `rustfmt` changed quite a bit in 0.7.0. (See failed CI.)

In [hbbft](https://github.com/poanetwork/hbbft/blob/master/.travis.yml), we fixed the Rust version in Travis to one particular nightly. I don't think `cargo +stable fmt` actually uses a stable `rustfmt` version: On my machine that still uses the new 0.7.0.

I'm happy to apply the new formatting and/or fix a nightly version. Let me know which you prefer.

(Also note that with rustfmt 0.7.0, `--write-mode` has been replaced by `--check`, so that needs to be updated in the Travis file, too. Vim users need to use `let g:rustfmt_emit_files = 1`, and this branch of rust.vim until it's merged: https://github.com/rust-lang/rust.vim/pull/213)

afck:
@c0gent verified that it builds on windows-msvc.

The CI failure is probably because of Rust 1.28.0: Since `.travis.yml` just uses `stable`, this is likely to break every six weeks, I think.

PR #127, opened 2018-05-28
Re-format using latest rustfmt.

tomusdrw:


rustfmt.toml
```diff
@@ -1 +1,3 @@
-max_width = 600
+max_width = 1000
```
afck: That's very… unorthodox.
Wouldn't it be better to break all those long string literals into shorter lines instead and make the width `100` (that's default anyway, I think)?

PR #129, opened 2018-06-04
add logs function to Eth (work by debris)

snd:
this has been lying around unmerged for months.

seems to be a fairly small change. i think we should merge this.

feel free to close this with a reason.

@debris @tomusdrw


tomusdrw:
Seems like a duplicate of a bit more complete #122

snd:
indeed!

PR #136, opened 2018-07-04
Set Block seal_fields attribute to an empty vector if missing(#135)

### LINKED ISSUE #135: Block serialization with Serde fails if sealFields is missing
`sealFields` appears to not be supported by all Ethereum implementations. One such example is Ganache. The problem is that the way Serde is set up for the `Block` type makes requests to those implementations panic.

My recommended solution is to serialize `sealFields` to an empty vector if it is not included in the response to e.g. `eth_getBlockByHash`. I'll provide a pull request shortly.

Jannis:
This fixes #135. By using `#[serde(default, ...)]` the value of `seal_fields` is set to an empty vector if it is not present in a block.

### DIFF
--- a/src/types/block.rs
+++ b/src/types/block.rs
@@ -90,7 +90,7 @@ pub struct Block<TX> {
     #[serde(rename = "totalDifficulty")]
     pub total_difficulty: U256,
     /// Seal fields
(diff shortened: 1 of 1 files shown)


tomusdrw:
Looks good, thanks!

PR #148, opened 2018-08-16
Add `traces` namespace

InsidiousMind:
So I'm currently doing a project that makes heavy use of parity's `trace` API. I was using web3 for it, creating a custom `namespace` to get the job done. However, I was coding up most of the API so i figured to just send it in as a PR and see what happens.

I'm not sure if this is exactly what your looking for, since this is a parity specific `traces` namespace (I wasn't sure how to integrate other RPC's, since Geth uses a different method altogether (`debug_trace` and not `trace_`). I was thinking that could be it's own namespace (ie debug()) or in traces as (`debug_(geth method)`).


src/api/traces.rs
```diff
@@ -0,0 +1,231 @@
...
+        let filter = helpers::serialize(&filter);
+        CallResult::new(self.transport.execute("trace_filter", vec![filter]))
+    }
+
+    /// Returns trace at the given position
+    pub fn get(&self, hash: H256, index: Vec<Index>) -> CallResult<Trace, T::Out> {
+        let hash = helpers::serialize(&hash);
+        let index = helpers::serialize(&index);
+        CallResult::new(self.transport.execute("trace_get", vec![hash, index]))
+    }
```
tomusdrw: Missing blank line

src/api/traces.rs
```diff
@@ -0,0 +1,231 @@
...
+}
+
+#[cfg(test)]
+mod tests  {
+    use futures::Future;
+
+    use api::Namespace;
+    use types::{TraceType, BlockNumber, BlockTrace, Trace, TraceFilterBuilder, Bytes, CallRequest, H256};
+
+    use super::Traces;
```
tomusdrw: Would add a blank line as well.

src/types/log.rs
```diff
@@ -93,6 +74,25 @@ pub struct Filter {
     limit: Option<usize>,
 }
 
+#[derive(Default, Debug, PartialEq, Clone)]
```
tomusdrw: Why was it moved? I think I preferred when `Filter` and `FilterBuilder` were next to each other.

src/types/trace_filtering.rs
```diff
@@ -0,0 +1,465 @@
...
+    /// Block Hash
+    pub block_hash: H256,
+}
+
+macro_rules! de_value {
+    ($action: ident) => ({
+        serde_json::from_value($action).map_err(|e| de::Error::custom(e.to_string()))
+    })
+}
+// a pretty standard custom deserialize, except it deserializes 'error' and 'result' of JSON
```
tomusdrw: Is it necessary? Can't we use serde's tagged representation here? Seems like a lot of code to maintain.

src/types/trace_filtering.rs
```diff
@@ -0,0 +1,465 @@
...
+// into the result enum, as well as deserializes `Action` based upon `type` field of the JSON.
+impl<'de> Deserialize<'de> for Trace {
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error> where D: Deserializer<'de> {
```
tomusdrw: Please add some more tests to this deserializer, we should handle possible error cases as well, cause the stringy code seems quite fragile.

src/types/traces.rs
```diff
@@ -0,0 +1,300 @@
...
+	trace_address: Vec<Address>,
+	/// Subtraces
+	subtraces: usize,
+	/// Action
+	action: Action,
+	/// Result
+	result: Option<Res>,
+}
+
+macro_rules! de_value {
```
tomusdrw: Could be a generic function?

src/types/traces.rs
```diff
@@ -0,0 +1,300 @@
...
+                };
+                let result = result.ok_or_else(|| de::Error::missing_field("result"))?;
+                let trace_address = trace_address.ok_or_else(|| de::Error::missing_field("trace_address"))?;
+                let subtraces = subtraces.ok_or_else(|| de::Error::missing_field("subtraces"))?;
+
+                Ok( TransactionTrace { trace_address, subtraces, action, result } )
+            }
+        }
+
+        const FIELDS: &'static [&'static str] = &["subtraces", "trace_address", "action", "type", "error", "result" ];
```
tomusdrw: Same as with other deserialize, can't we use serde-derived deserialization?
Maybe add a helper struct that matches the JSON format (with derived Deserialization) and then custom deserializer for this one would be just converting one representation to another?

InsidiousMind:
Hey!

Just wanted to give an update for where I am at with this. My semester has started ~ a week ago, so I won't be able to work on this as much. I plan to get around to it as soon as I can, but I am mostly working on my Senior Thesis Project, which is due in ~two months.

Just wanted to leave this here in case your wondering what might be taking me so long ;)

tomusdrw:
Hi @InsidiousMind thank you for the heads-up, we don't have a deadline for this one :) Good luck with your project!

InsidiousMind:
I removed the custom deserializers for TraceTransaction and Trace, because serde's `untagged` representation worked well. I made sure of it by writing some more tests for the deserialize in `trace_filtering`

Issue #116, opened 2018-04-24
Signing Transactions Offline


emkman:
Support for signing transactions offline for sending to a locked node or potentially Infura would be extremely helpful. Similar functionality can be found in the `ethereumjs-tx` for javascript. Much of this functionality exists across the parity codebase and needs to be pieced together. In discussing this with @tomusdrw, the general steps required would be:

1.  Encode contract call (this can be done by `ethabi`)
2. Create and fill a transaction request (nonce, gas, gasprice, etc) - I believe there is an RPC method to do that in Parity
3. Encode the transaction into RLP
4. Calculate hash and sign (this could be done by `ethkey` library)
5. Broadcast signed transaction

synlestidae:
Been interested in having this functionality since I came across the web3 crate. 

I propose we implement `web3::api::Eth::sign_transaction`:

`
pub fn sign_transaction(&self, transaction: &Transaction, private_key: H256) -> CallResult<Bytes, T::Out> { 
   // ...
}
`

As Transaction has redundant fields like `from` (derived from private key) and `hash` (computed from transaction), we could use a new struct. `UnsignedTransaction` perhaps.

As I understand it, Web3 just needs to call [signTransaction](https://web3js.readthedocs.io/en/1.0/web3-eth-accounts.html#signtransaction) with the right parameters. It'll take a bit of learning but I'd like to take this task :)

synlestidae:
The output of `sign_transaction` would be the RLP-encoded transaction that can be fed into `send_raw_transaction`.

tomusdrw:
Most likely integrate with https://github.com/tomusdrw/ethsign

steveswing:
> Most likely integrate with https://github.com/tomusdrw/ethsign

We used ethsign with web3 indirectly -- in https://github.com/SubstratumNetwork/SubstratumNode -- but had to copy RawTransaction to work around type conflicts between web3::types, ethereum-types, and primitive-types.

tomusdrw:
@steveswing cool! Yeah, there is no proper transaction type around (except for parity repo, but you really don't want to reference that in your code). I also wanted to avoid it in ethsign (initially), but would be happpy to include it at some point, along with different signature payloads that Ethereum is using (signData, signTransaction, etc)).

PR #152, opened 2018-08-25
WIP: RawTransaction struct for signing offline (#116)

synlestidae:
So I made this but I don't think it's quite right. For a start, it uses three new dependencies for transaction signing to work offline.

Thoughts?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,8 +23,11 @@ rustc-hex = "1.0"
 serde = "1.0"
 serde_derive = "1.0"
 serde_json = "1.0"
+tiny-keccak="1.4.2"
+secp256k1 = "0.11.1"
 tokio-timer = "0.1"
 url = "1.7"
+rlp = "0.2.2"
 # Optional deps
 hyper = { version = "0.11", optional = true }
 hyper-tls = { version = "0.1", optional = true }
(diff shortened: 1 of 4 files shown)


tomusdrw:
Have a look at `ethkey`: https://github.com/paritytech/parity-ethereum/tree/master/ethkey might be worth using this crate (I believe it's published on crates.io) to handle private keys and signing.

We should also handle `chainId` in the signature for replay-protection (currently setting `v` to a fixed value is incorrect in your code).

Offline-signing should also be feature-gated, so that we can disable that if not needed.

I'm starting to think that maybe entire transaction-signing should be handled by a separate library (like `ethereum-tx-signer`) and rust-web3 would only use that one, cause there is quite a lot of logic involved in that and it might be useful outside of rust-web3 context as well.

synlestidae:
I also thought this issue could make sense as a separate crate since the changes here don't seem to be congruent with the rest of the rust-web3 crate. You can now find these changes at https://crates.io/crates/ethereum-tx-sign. It's still a WIP and needs more testing and feedback. It is now compatible with EIP-155, which sorted out the ChainId issue.

tomusdrw:
Awesome, if we keep `ethereum-types` in sync the crate should also be easy to integrate with `rust-web3`, I'm happy to include usage examples in this repo.

PR #161, opened 2018-10-09
Ensure that 'to' = None when it is a contract creation transaction

D4nte:
Good day,

The documentation on the Transaction struct says that `Transaction.to` is "Recipient (None when contract creation)".

However, in web3 json response, the `to` field actually contains address zero (used for contract deployments), hence this field is not always `None`. Tested with ganache & infura.

This change is to ensure that if the `to` field is present and it is zero address, then `None` is used.

Let me know if you prefer me to implement this in a different manner, I could not think of something more elegant.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -46,5 +46,6 @@ http = ["hyper", "tokio-core"]
 ipc = ["tokio-uds", "tokio-core", "tokio-io"]
 ws = ["tokio-core", "websocket"]
 tls = ["hyper-tls", "native-tls"]
+account-zero = []
 
 [workspace]
--- a/src/types/transaction.rs
+++ b/src/types/transaction.rs
@@ -1,4 +1,6 @@
 use types::{Bytes, H160, H256, Index, Log, U256, U64};
+use serde::{de, export::fmt, Deserializer};
+use serde_json;
 
 /// Description of a Transaction, pending or in the chain.
 #[derive(Debug, Default, Clone, PartialEq, Deserialize, Serialize)]
@@ -19,6 +21,7 @@ pub struct Transaction {
     /// Sender
     pub from: H160,
     /// Recipient (None when contract creation)
+    #[serde(deserialize_with = "deserialize_to")]
     pub to: Option<H160>,
     /// Transfered value
     pub value: U256,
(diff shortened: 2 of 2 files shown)


D4nte:
Thanks for the feedback, I am raising the issue to ganache and infura.
I aim to move to parity for testing purposes, hopefully parity behaves correctly.

thomaseizinger:
As far as I understand it, you distinguish exactly by that: if the transaction goes to the account `0` it is a contract creation. I couldn't find it in the yellow paper but these docs here say so:

https://solidity.readthedocs.io/en/develop/introduction-to-smart-contracts.html#index-8

Here is a Stackoverflow post that explains it in more detail: https://stackoverflow.com/a/48220805

tomusdrw:
@thomaseizinger Zero address is here: https://etherscan.io/address/0x0000000000000000000000000000000000000000 and it's a valid address that can be recipient of the transaction.

The docs you've linked are quite old, I actually made a PR to update them earlier today
https://github.com/ethereum/solidity/pull/5174

In the stack overflow thread you can even find a discussion, were people are confused with what `address(0)` actually is.

On the RPC level the distinction between contract creation and transfer is done by checking `null` value (or presence) of the `to` field. The less-technical docs are however using phrases like "empty account" which may be interpreted as zero-address, especially backed up by the incorrect solidity documentation.

D4nte:
I have engaged ganache devs regarding the issue. https://gitter.im/ConsenSys/truffle?at=5bbc4af7c08b8b3067f1c674

After checking the web3 api, I agree with you, it is written that it should return `null`. It is unfortunate that these web3 implementations do not comply with the expected standard.

However, is there a compromise you would consider to make so that the rust-web3 library can be used with these various implementations?

tomusdrw:
@D4nte Happy to accept the change hidden by opt-in cargo feature, to allow (hopefuly temporary) compatibility with ganache / infura.
BTW my PR to update docs got accepted :yay:, so hopefuly we can straighten up the understanding of this :)

D4nte:
The issue is actually only with ganache. I hide the change behind a cargo feature.
Thank you for the doc update, I'll mention them in the ganache issue.

D4nte:
This code is not working + ganache has accepted my issue so let's close it.

D4nte:
See https://github.com/trufflesuite/ganache-cli/issues/579

PR #170, opened 2018-11-10
Fix build with OpenSSL 1.1.1

### LINKED ISSUE #162: openssl 1.1.1 - can't build  `openssl v0.9.24`
See https://github.com/sfackler/rust-openssl/issues/994 for details

All crates that are using `openssl v.0.9.24` need be bumped/patched such as:
* `jsonrpc-core`
* `hyper`
* `hyper-tls`
* `native-tls`
* `websocket`

InsidiousMind:
Update hyper to 0.12
Update hyper-tls to 0.3
update websocket to 0.21
Update http.rs to new hyper 0.12 API
bump version

Fixes #162 

This fix means upgrading to hyper 0.12 to support OpenSSL 1.1.1. This works out OK, but the largest change to take note of would probably be the removal of typed-headers in 0.12.

### DIFF
--- a/src/transports/http.rs
+++ b/src/transports/http.rs
@@ -167,10 +167,10 @@ impl Http {
         if len < MAX_SINGLE_CHUNK {
             req.headers_mut().insert(hyper::header::CONTENT_LENGTH, len.into());
         }
+        // Send basic auth header
         if let Some(ref basic_auth) = self.basic_auth {
             req.headers_mut().insert(hyper::header::AUTHORIZATION, basic_auth.clone());
         }
-        // Send basic auth header
         let (tx, rx) = futures::oneshot();
         let result = self.write_sender
             .unbounded_send((req, tx))
--- a/src/transports/http.rs
+++ b/src/transports/http.rs
@@ -14,9 +14,11 @@ use std::sync::atomic::{self, AtomicUsize};
 
 use futures::sync::{mpsc, oneshot};
 use futures::{self, future, Future, Stream};
+use self::hyper::header::HeaderValue;
 use helpers;
 use rpc;
 use serde_json;
+use base64;
 use transports::Result;
 use transports::shared::{EventLoopHandle, Response};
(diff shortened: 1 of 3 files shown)


src/transports/http.rs
```diff
@@ -119,15 +126,17 @@ impl Http {
...
             let user = url.username();
 
             if user.len() > 0 {
-                Some(hyper::header::Basic {
-                    username: user.into(),
-                    password: url.password().map(Into::into),
-                })
+                let auth = match url.password() {
+                    Some(pass) => format!("{}:{}", user, pass),
+                    None => format!("{}:", user)
```
tomusdrw: you sure we need the trailing `:`? Maybe let's encode only the username.

src/transports/http.rs
```diff
@@ -147,24 +156,21 @@ impl Http {
...
         }
-        // Send basic auth header
         if let Some(ref basic_auth) = self.basic_auth {
-            req.headers_mut()
-                .set(hyper::header::Authorization(basic_auth.clone()));
+            req.headers_mut().insert(hyper::header::AUTHORIZATION, basic_auth.clone());
         }
-        req.set_body(request);
-
+        // Send basic auth header
```
tomusdrw: The comment should be above `if` I suppose :)

src/transports/http.rs
InsidiousMind: I was conflicted on this myself, since the HTTP Basic Auth RFC does not specify exactly what to do with an empty password, just that the format must be 'Basic username:password', with user:pass part encoded in Base64. I am not well-versed on HTTP spec, though, so I may be wrong on this, but figured it would be safest to do 'user:'

https://tools.ietf.org/html/rfc7617#section-2

My conclusion was no password just counts as an 'empty' password so it would be encoded as 'Basic username:'

tomusdrw: Ok, makes sense!

PR #173, opened 2018-11-27
Increased the number of possible outputs from a web3 query

AvishaiW:
Hi,
Until now it was only possible to retrieve 5 arguments as a return value from a query function because `Detokenize` was implemented only up to a tuple of 5.
I modified it so it will match the number of possible inputs - 16.

It's not the most efficient way to do it, but I guess that's a deeper problem to solve.

### DIFF
--- a/src/contract/tokens.rs
+++ b/src/contract/tokens.rs
@@ -60,7 +60,17 @@ impl_output!(2, A, B,);
 impl_output!(3, A, B, C,);
 impl_output!(4, A, B, C, D,);
 impl_output!(5, A, B, C, D, E,);
-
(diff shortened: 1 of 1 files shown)


elichai:
Hey @tomusdrw this looks pretty important. what do you think?

tomusdrw:
Thanks!

tomusdrw:
@AvishaiW @elichai If you guys need this released soon, I'd be really grateful for a PR bumping the version to 0.5.1, then I can release to creates.io

elichai:
@tomusdrw Thanks! 
Here: https://github.com/tomusdrw/rust-web3/pull/174

Issue #180, opened 2019-01-03
mixHash and nonce in the BlockHeader struct


fredfortier:
Hi - What's the reason for omitting `mixHash` and `nonce` in the BlockHeader struct?

tomusdrw:
@fredfortier No reason at all, just forgotten. I was basing the fields on the Parity Ethereum implementation, and `mixHash` and `nonce` are added in a different way than other fields so I didn't notice them.
PRs welcome

fredfortier:
Ok good to know! I’ll create a PR today.

fredfortier:
@tomusdrw Here is some background on the issue from Parity: https://github.com/paritytech/parity-ethereum/issues/8841.

Unless you suggest otherwise, I'll submit a PR that adds these fields to the `Block` and `BlockHeader` struct as `Option<T>`. Some chains (POA chains) do not return them. Geth used a similar approach: https://github.com/ethereum/go-ethereum/pull/18166/commits

Web3.js does not cast these fields: https://github.com/ethereum/web3.js/blob/b4c1542ddb5284267f6814ba0106bfbbc83fe166/test/web3.eth.getBlock.js but they are returned by their JSON-RPC passthrough:
```
web3.eth.getBlock("0x04b6d5b1801fff93df4d35c828202392055bcb630d656fc98f74410427344cde");
{ number: 13,
  hash: '0x04b6d5b1801fff93df4d35c828202392055bcb630d656fc98f74410427344cde',
  parentHash: '0x5013631d38bd1b9924e56e2e23106dedb8c4cea8786ac879d2ad0f2d7253f4c9',
  mixHash: '0x1010101010101010101010101010101010101010101010101010101010101010',
  nonce: '0x0000000000000000',
  sha3Uncles: '0x1dcc4de8dec75d7aab85b567b6ccd41ad312451b948a7413f0a142fd40d49347',
  logsBloom: '0x00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000002000000000001000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000002000000000000000000000000000000000000000000000000000000040000000000000000000000000000000000000000002000000000000000000000000000000000000000000000000000000000000000000000000800000000000000',
  transactionsRoot: '0x56e81f171bcc55a6ff8345e692c0f86e5b48e01b996cadc001622fb5e363b421',
[...]

PR #181, opened 2019-01-07
Added Attributes to Receipt, Block and BlockHeader Structs

fredfortier:
As discussed in issue #180 , I added optional `mixHash` and `nonce` attributes to the Block and BlockHeader structs. In addition, I added `logsBloom` to the Receipt struct.


src/types/block.rs
```diff
@@ -98,6 +103,10 @@ pub struct Block<TX> {
     pub transactions: Vec<TX>,
     /// Size in bytes
     pub size: Option<U256>,
+    #[serde(rename = "mixHash")]
```
tomusdrw: Missing docs?

fredfortier: Sorry, I'll adjust in a minute.

fredfortier: Added the missing docstring and changed the type of `nonce` to H64. Here is a reference that supports H64 as the correct type: https://github.com/ethereum/go-ethereum/blob/461291882edce0ac4a28f64c4e8725b7f57cbeae/core/types/block.go:43. Feel free to double check.

fredfortier: Fixing the unit test that I broke...

src/api/eth.rs
```diff
@@ -416,6 +418,7 @@ mod tests {
     "cumulativeGasUsed": "0x33bc",
     "gasUsed": "0x4dc",
     "contractAddress": "0xb60e8dd61c5d32be8058bb8eb970870f07233155",
```
fredfortier: Your `EXAMPLE_RECEIPT` states that it was taken from the Ethereum JSON-RPC docs but did not include `logsBloom`. I verified the docs: https://github.com/ethereum/wiki/wiki/JSON-RPC#eth_gettransactionreceipt ; I do see a value for `logsBloom`, so I included it. I'm not sure if there was a specific reason for excluding it. All tests are now passing on my side.

PR #183, opened 2019-01-08
Set gas limit and gas in contract deploy

jedahan:
This is a bugreport mostly
I have to manunally set the gas & gas limit to get past "base fee exceeds limit"

But now I am stuck with a revert.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -34,6 +34,7 @@ tokio-core = { version = "0.1", optional = true }
 tokio-io = { version = "0.1", optional = true }
 tokio-uds = { version = "0.1", optional = true }
 websocket = { version = "0.21", optional = true }
+env_logger = "0.6.0"
 
 [dev-dependencies]
 # For examples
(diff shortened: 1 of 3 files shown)


examples/readme.md
```diff
@@ -0,0 +1,13 @@
+First, build the examples
+
+    cargo build --examples
```
tomusdrw: This is redundant, `cargo run` does the compilation as well

examples/contract.rs
```diff
@@ -20,7 +22,9 @@ fn main() {
         .unwrap()
         .confirmations(4)
         .options(Options::with(|opt| {
-            opt.value = Some(5.into())
+            opt.value = Some(5.into());
+            opt.gas_price = Some(5.into());
+            opt.gas = Some(1_000_000.into());
```
tomusdrw: `gas` is optional in the RPC - backend should figure out the value itself. I suspect that your transaction always reverts hence ganache is not able to estimate how much gas is needed (it reverts no matter how much gas it provides so it eventually fails by reaching block gas limit)

Cargo.toml
```diff
@@ -34,6 +34,7 @@ tokio-core = { version = "0.1", optional = true }
 tokio-io = { version = "0.1", optional = true }
 tokio-uds = { version = "0.1", optional = true }
 websocket = { version = "0.21", optional = true }
+env_logger = "0.6.0"
```
tomusdrw: Should go to `dev-dependencies`

PR #187, opened 2019-01-27
Implement personal_signTransaction

folex:
I've introduced a new type `SignedTransaction` which has just a single field `raw: Bytes`. Both Parity and Geth also return `tx` field in response, but I ignored it because it's different between Parity and Geth.

*Parity 2.2.7 (stable)* response looks like this (it's from docs, but I tested it, it looks like this):
```json
{
  "id": 1,
  "jsonrpc": "2.0",
  "result": {
    "raw": "..",
    "tx": {
      "hash": "0xc6ef2fc5426d6ad6fd9e2a26abeab0aa2411b7ab17f30a99d3cb96aed1d1055b",
      "nonce": "0x0",
      "blockHash": "0xbeab0aa2411b7ab17f30a99d3cb9c6ef2fc5426d6ad6fd9e2a26a6aed1d1055b",
      "blockNumber": "0x15df",
      "transactionIndex": "0x1",
      "from": "0x407d73d8a49eeb85d32cf465507dd71d507100c1",
      "to": "0x853f43d8a49eeb85d32cf465507dd71d507100c1",
      "value": "0x7f110",
      "gas": "0x7f110",
      "gasPrice": "0x09184e72a000",
      "input": "0x603880600c6000396000f300603880600c6000396000f3603880600c6000396000f360"
    }
  }
}
```

*Geth 1.8.22 (unstable)* response looks like this:
```json
{
    "jsonrpc": "2.0",
    "id": 1,
    "result": {
        "raw": "..",
        "tx": {
            "nonce": "0x0",
            "gasPrice": "0xffff",
            "gas": "0xf4240",
            "to": "0x9a3027b366a50b3b451811e9b4cf48b4f96d780e",
            "value": "0xffff",
            "input": "0x",
            "v": "0x9c8",
            "r": "0x4a3a70f717b8a9a5f2fc2ab2814d00d61a11f7678047b8cd6bbe402be56bf316",
            "s": "0x588c81b447319cd164cc8717311027a0f032f1bb405b94122a817b681fab7557",
            "hash": "0xc46996c2c0d291266c24d21b2abe1de2150a3a59f5182d98e57f868b50abd79e"
        }
    }
}
```

The main reason is that Geth misses `from` which is a non-optional field in `web3::transaction::Transaction`, so straightforward deserialization isn't possible. 

Maybe there is some workaround since we actually always know `from`, and maybe there is a way to make `serde` use it. But I don't really see a point in that, since sender already have all that info in `TransactionRequest`.

That said, maybe there is a point in returning `Bytes` directly, avoiding `SignedTransaction` wrapper altogether. But I'm not sure of an idiomatic way to do that, so I could use advice if you think that's the right way.

### DIFF
--- a/src/api/personal.rs
+++ b/src/api/personal.rs
@@ -2,7 +2,7 @@
 
 use api::Namespace;
 use helpers::{self, CallFuture};
-use types::{Address, H256, TransactionRequest};
+use types::{Address, H256, TransactionRequest, SignedTransaction};
 
 use Transport;
 
@@ -63,6 +63,18 @@ impl<T: Transport> Personal<T> {
                 .execute("personal_sendTransaction", vec![transaction, password]),
         )
     }
+
+    /// Signs a transaction without dispatching it to the network.
+    /// The account does not need to be unlocked to make this call, and will not be left unlocked after.
+    /// Returns a signed transaction in raw bytes along with details.
+    pub fn sign_transaction(&self, transaction: TransactionRequest, password: &str) -> CallFuture<SignedTransaction, T::Out> {
+        let transaction = helpers::serialize(&transaction);
+        let password = helpers::serialize(&password);
+        CallFuture::new(
+            self.transport
+                .execute("personal_signTransaction", vec![transaction, password]),
+        )
+    }
 }
 
 #[cfg(test)]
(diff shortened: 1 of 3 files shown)


folex:
Just to be certain, you mean add a separate type to handle `result.tx` in JSON? Ok, sure. 

I'm confused by name suggestion of `Transaction`. There is already a `web3::transaction::Transaction`, so to avoid name clashing and confusion, I'd name the new, unfilled transaction type as something like `CrudeTransaction` or `BlankTransaction`. What do you think?

tomusdrw:
> Just to be certain, you mean add a separate type to handle result.tx in JSON?

Yes, but only if you feel like it, not required to get that PR merged (just let me know if you want to work on this more)

> I'd name the new, unfilled transaction type as something like CrudeTransaction or BlankTransaction. What do you think?

I'd keep `web3::transaction::Transaction`, rename `SignedTransaction` to `RawTransaction` and the new type would be called `RawTransactionDetails`

folex:
Yep, I would definitely like to add transaction details, to `SignedTransaction`. 

Just to make my previous message clear: I would definitely keep `web3::transaction::Transaction`, didn't mean to rename it at all! :) I was meaning to name type of `SignedTransaction::tx` something like `BlankTransaction`.

Ok, I'll go with `RawTransaction` and `RawTransactionDetails` then, thanks for help with naming! I think I'll commit changes today or tomorrow.

folex:
Looks good?

PR #189, opened 2019-01-28
Make TransactionReceipt::gas_used optional

folex:
I'm using Parity 2.3.0 in `--light` mode on Kovan. And I'm seeing the following response when asking for a transaction receipt:
```json
{
  "jsonrpc": "2.0",
  "result": {
    "blockHash": "0x56c4aae4ef18eccd3fa93be10c0fbf7625ff96d2b1f842cf2e37f03a52758fa4",
    "blockNumber": "0x9bdc13",
    "contractAddress": null,
    "cumulativeGasUsed": "0x4b018",
    "from": null,
    "gasUsed": null,
    "logs": [],
    "logsBloom": "0x00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000",
    "root": null,
    "status": "0x0",
    "to": null,
    "transactionHash": "0xd9447dc668773052f1967ff2a2401f41d49ad5a5d5de6b1a9fb482dc98bec270",
    "transactionIndex": "0x2"
  },
  "id": 1
}
```

`gasUsed` is null, and I'm getting a panick from web3, since `gas_used` is a non-optional field, so I fixed this by making it optional.

But I'm concerned about this being a breaking change to `rust-web3` API. Maybe I should use default value here?


src/types/transaction.rs
```diff
@@ -51,7 +51,7 @@ pub struct Receipt {
     pub cumulative_gas_used: U256,
     /// Gas used by this transaction alone.
```
tomusdrw:
```suggestion
    /// Gas used by this transaction alone.
    ///
    /// Gas used is `None` if the the client is running in light client mode.
```

folex:
I bumped minor version to `0.5.2` :)

PR #194, opened 2019-02-12
added peer listing

blanksteer:
query peer data

### DIFF
--- a/src/api/net.rs
+++ b/src/api/net.rs
@@ -37,8 +37,8 @@ impl<T: Transport> Net<T> {
     }
 
     /// Returns list of connected nodes
-    pub fn peer_list(&self, _n: NodeType) -> CallFuture<PeerType, T::Out>{
-        match _n {
+    pub fn peer_list(&self, n: NodeType) -> CallFuture<PeerType, T::Out>{
+        match n {
             NodeType::Parity => CallFuture::new(self.transport.execute("parity_netPeers", vec![])),
             NodeType::Geth => CallFuture::new(self.transport.execute("admin_peers", vec![])),
         }
--- a/src/api/net.rs
+++ b/src/api/net.rs
@@ -2,7 +2,7 @@
 
 use api::Namespace;
 use helpers::CallFuture;
-use types::U256;
+use types::{U256, PeerType, NodeType};
 
 use Transport;
 
@@ -34,6 +34,14 @@ impl<T: Transport> Net<T> {
     /// Returns number of peers connected to node.
     pub fn peer_count(&self) -> CallFuture<U256, T::Out> {
         CallFuture::new(self.transport.execute("net_peerCount", vec![]))
+    }
+
+    /// Returns list of connected nodes
+    pub fn peer_list(&self, _n: NodeType) -> CallFuture<PeerType, T::Out>{
+        match _n {
+            NodeType::Parity => CallFuture::new(self.transport.execute("parity_netPeers", vec![])),
+            NodeType::Geth => CallFuture::new(self.transport.execute("admin_peers", vec![])),
+        }
     }
 
     /// Whether the node is listening for network connections
--- a/src/types/node.rs
+++ b/src/types/node.rs
@@ -1,0 +1,6 @@
+//! Type of node (Parity, Geth...)
+
+pub enum NodeType {
+    Parity,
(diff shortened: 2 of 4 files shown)


src/api/net.rs
```diff
@@ -36,6 +36,14 @@ impl<T: Transport> Net<T> {
         CallFuture::new(self.transport.execute("net_peerCount", vec![]))
     }
 
+    /// Returns list of connected nodes
+    pub fn peer_list(&self, _n: NodeType) -> CallFuture<PeerType, T::Out>{
```
tomusdrw: use `_` prefix only for unused parameters

tomusdrw: While I don't fully agree with placing it in the `Net` namespace, I don't have a good alternative to this. Let's leave it here then :)

tomusdrw:
Seems that `node.rs` file is missing?

src/types/node.rs
```diff
@@ -0,0 +1,6 @@
+//! Type of node (Parity, Geth...)
+
+pub enum NodeType {
```
tomusdrw: Would you mind adding documentation to all public types and variants in this file and `peer.rs`?

src/types/peer.rs
```diff
@@ -0,0 +1,66 @@
...
+    pub active: usize,
+    pub connected: usize,
+    pub max: u32,
+    pub peers: Vec<PeerInfo>,
+}
+
+impl fmt::Debug for PeerType {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(f, "active: {}, connected: {}, max: {}, peers: {:?}", self.active, self.connected, self.max, self.peers)
+
```
tomusdrw: stray newline

src/types/peer.rs
```diff
@@ -0,0 +1,66 @@
...
+#[derive(Serialize, Deserialize, Clone)]
+pub struct PeerType {
+    pub active: usize,
+    pub connected: usize,
+    pub max: u32,
+    pub peers: Vec<PeerInfo>,
+}
+
+impl fmt::Debug for PeerType {
```
tomusdrw: Why not `#[derive(Debug)]`?

src/types/peer.rs
```diff
@@ -0,0 +1,66 @@
...
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(f, "id: {:?}, name: {}, caps: {:?}, remote address: {}, local address: {}", self.id, self.name, self.caps, self.network.remote_address, self.network.local_address)
+    }
+}
+
+
+#[derive(Serialize, Clone, Deserialize, Debug)]
+pub struct PeerNetworkInfo {
+
+    #[serde(rename="remoteAddress")]
```
tomusdrw: could use `#[serde(rename_all = "camelCase")]` on the entire struct

src/api/net.rs
blanksteer: Fixed param prefix. Wasn't sure about putting it in it's own namespace. Saw that peer_count was under Net. Will leave it under Net until told otherwise.

src/types/peer.rs
blanksteer: older version of nightly gave error that Debug needed to be impl for PeerType. upgraded to latest nightly and error went away. switched to ``#[derive(Debug)]`` and removed debug impl.

src/types/peer.rs
blanksteer: done.

src/types/peer.rs
blanksteer: stray newline should be fixed now.

src/types/node.rs
blanksteer: added missing docs to both ``node.rs`` and ``peer.rs``.

tomusdrw:
Thanks!

blanksteer:
@tomusdrw i was so excited that you considered this idea that i forgot to discuss an issue! the json returned from a parity node vs a geth node are completely different. I would like to either setup two different funcs in the ``Net`` namespace with it's appropriate return type or remove ``NodeType`` along with it's match statement in the current ``peer_list`` function. This second option would only allow deserializing from a parity node. I await your thoughts/direction on this issue.

PR #196, opened 2019-02-13
parity_netPeers function

blanksteer:
geth returns a different structure. believe it is of type HashMap. thought this would be a simpler/better alternative atm until i have time to do some digging through geth's source. will setup a seperate func for geth. hope the naming scheme is ok. ``func parity_peer_list`` and ``func geth_peer_list``.

### DIFF
--- a/src/api/net.rs
+++ b/src/api/net.rs
@@ -37,8 +37,8 @@ impl<T: Transport> Net<T> {
     }
 
     /// Returns list of connected nodes
-    pub fn peer_list(&self, _n: NodeType) -> CallFuture<PeerType, T::Out>{
-        match _n {
+    pub fn peer_list(&self, n: NodeType) -> CallFuture<PeerType, T::Out>{
+        match n {
             NodeType::Parity => CallFuture::new(self.transport.execute("parity_netPeers", vec![])),
             NodeType::Geth => CallFuture::new(self.transport.execute("admin_peers", vec![])),
         }
--- a/src/api/net.rs
+++ b/src/api/net.rs
@@ -2,7 +2,7 @@
 
 use api::Namespace;
 use helpers::CallFuture;
-use types::U256;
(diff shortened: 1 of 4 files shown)


src/api/net.rs
```diff
@@ -36,6 +36,11 @@ impl<T: Transport> Net<T> {
         CallFuture::new(self.transport.execute("net_peerCount", vec![]))
     }
 
+    /// Returns list of connected nodes parity
+    pub fn parity_peer_list(&self) -> CallFuture<PeerType, T::Out>{
```
tomusdrw: Let's move this to `parity_` namespace then

src/types/node.rs
```diff
@@ -0,0 +1,9 @@
+//! Type of node (Parity, Geth...)
+
+/// specify behavior of node interacting with
+pub enum NodeType {
```
tomusdrw: This doesn't seem required now

src/types/peer.rs
```diff
@@ -0,0 +1,65 @@
...
+/// ip address of both local and remote
+#[derive(Debug, Serialize, Deserialize, Clone)]
+#[serde(rename_all="camelCase")]
+pub struct PeerNetworkInfo {
+    pub remote_address: String,
+    pub local_address: String,
+}
+
+#[derive(Debug, Serialize, Deserialize, Clone)]
+pub struct PeerProtocolsInfo {
```
tomusdrw: More docs

src/api/net.rs
blanksteer: will do! :) I think it's ``ParitySet``? Didn't realize you gave these instructions. Asked if i had the go ahead on [#198](https://github.com/tomusdrw/rust-web3/pull/198)

tomusdrw: Sounds good!

PR #199, opened 2019-03-04
Parity net peers

blanksteer:
added parity_netPeers to ParitySet namespace. function gives list of connected/connecting nodes.


src/api/parity_set.rs
```diff
@@ -40,6 +39,11 @@ impl<T: Transport> ParitySet<T> {
         CallFuture::new(self.transport().execute("parity_dropNonReservedPeers", vec![]))
     }
 
+    /// Get list of connected/connecting peers.
+    pub fn parity_net_peers(&self) -> CallFuture<ParityPeerType, T::Out>{
```
tomusdrw: Could you add a test?

blanksteer: hopefully get to it today. :)

blanksteer:
sorry it took so long.

src/types/parity_peers.rs
```diff
@@ -18,8 +17,34 @@ pub struct ParityPeerType {
     pub peers: Vec<ParityPeerInfo>,
 }
 
+/// add functions for ParityPeerType
```
tomusdrw: All the helpers methods should be available only for tests (or rather just rewrite the test above as suggested and remove the helpers)

src/api/parity_set.rs
```diff
@@ -145,11 +144,16 @@ mod tests {
...
         ParitySet:add_reserved_peer,
-        "enode://a979fb575495b8d6db44f750317d0f4622bf4c2aa3365d6af7c284339968eef29b69ad0dce72a4d8db5ebb4968de0e3bec910127f134779fbcb0cb6d3331163c@22.99.55.44:7770" 
+        "enode://a979fb575495b8d6db44f750317d0f4622bf4c2aa3365d6af7c284339968eef29b69ad0dce72a4d8db5ebb4968de0e3bec910127f134779fbcb0cb6d3331163c@22.99.55.44:7770"
         => "parity_addReservedPeer", vec![r#""enode://a979fb575495b8d6db44f750317d0f4622bf4c2aa3365d6af7c284339968eef29b69ad0dce72a4d8db5ebb4968de0e3bec910127f134779fbcb0cb6d3331163c@22.99.55.44:7770""#];
         Value::Bool(true) => true
     );
 
+    rpc_test! (
+        ParitySet:parity_net_peers => "parity_netPeers";
+        serde_json::to_value(ParityPeerType::from(ParityPeerType::get_test_string())).unwrap() => ParityPeerType::from(ParityPeerType::get_test_string())
```
tomusdrw: We should rather have a `Value` on the left side and a hand-constructed struct on the right:
```rust
serde_json::from_str::<Value>(r#"{"active":5,"connected":5,"max":5,"peers":[{"id":"f900000000000000000000000000000000000000000000000000000000lalalaleelooooooooo","name":"","caps":[],"network":{"remoteAddress":"Handshake","localAddress":"127.0.0.1:43128"},"protocols":{"eth":null,"pip":null}}]}"#).unwrap() 
=> ParityPeerType {
  active: 5,
  connected: 5,
  max: 5,
  peers: vec![ParityPeerInfo {
     ...
  }],
}
```

The reason is to test deserialisation (from str), and the serialisation of hand-crafted type (to make sure we are able to instantiate the struct)
Basically we want to have the same stuff in two different formats (also for readability) to minimize the possibility of mistake.

blanksteer: thank you for explaining that. is it ok to leave everything pub for the time being or would you rather have getters/setters for ``ParityPeerType`` ?

tomusdrw: I'd prefer the types in this crate to be super simple and avoid having any logic. So keeping everything `pub` is definitely they way to go.

src/types/parity_peers.rs
```diff
@@ -17,14 +17,6 @@ pub struct ParityPeerType {
...
-/// add functions for ParityPeerType
-impl ParityPeerType {
-    /// string literal of json structure for a parity peer
-    pub fn get_test_string() -> &'static str {
-        r#"{"active":5,"connected":5,"max":5,"peers":[{"id":"f900000000000000000000000000000000000000000000000000000000lalalaleelooooooooo","name":"","caps":[],"network":{"remoteAddress":"Handshake","localAddress":"127.0.0.1:43128"},"protocols":{"eth":null,"pip":null}}]}"#
-    }
-}
-
 /// handle string literal conversion for tests
 impl From<&str> for ParityPeerType {
```
tomusdrw: This and below implementation should be removed as well.

Issue #205, opened 2019-03-25
Fix to edition 2018


botika:
The only dependence that gives error chain problems, waiting for https://github.com/rust-lang-nursery/error-chain/pull/255 .

tomusdrw:
We can get rid of `error-chain`, it's deprecated anyway, isn't it?

botika:
Totally agree. We get to this before migrating to Tokyo #146

vityafx:
Guys, who say that the `error-chain` crate is deprecated and why?

tomusdrw:
@vityafx it's definitely not maintained currently (see https://github.com/rust-lang-nursery/error-chain/pull/255#issuecomment-486126145) and uses stuff that is deprecated in the stdlib.

PR #206, opened 2019-03-25
Fix to edition 2018

botika:
#205

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,11 +14,15 @@ matrix:
   - rust: nightly
 
 before_script:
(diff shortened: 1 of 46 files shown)


botika:
I based myself on https://github.com/actix/actix-http/blob/master/src/error.rs#L206-L240.
I have to follow that I have a deadline. Is there a problem in the implementation? I continue from here with Tokyo?

sorry for the insistence, is that I need it.

tomusdrw:
Looks good, thanks a lot!

PR #209, opened 2019-03-31
Adding support for truffle's library linking

### LINKED ISSUE #186: Linked Libraries in Bytecode
I read every contract deployment example in this library. They all assume that the bytecode is a hex string that can be parsed into a byte array like this: `Contract::deploy(web3.eth(), include_bytes!("./build/SimpleEvent.abi"))`.

I'm pre-compiling my contracts with Truffle 5 and reading the bytecode form the `bytecode` attribute of the `build/contracts/MyContract.json` files. I'm observing two issues:
1. Truffle adds a `0x` prefix. This is easy, we just ignore the first two characters.
2. Truffle adds ascii library placeholders, like this: https://github.com/ethereum/solidity/issues/635. It appears that the solc-js compiler of Truffle uses the old method (fully qualified name of the library). However, based on this document: https://solidity.readthedocs.io/en/v0.5.3/using-the-compiler.html library placeholders are a Solidity compiler feature (not a Truffle quirk) and still use ascii characters: `If your contracts use libraries, you will notice that the bytecode contains substrings of the form __$53aea86b7d70b31448b230b20ae141a537$__. These are placeholders for the actual library addresses. The placeholder is a 34 character prefix of the hex encoding of the keccak256 hash of the fully qualified library name. `

Note that I made sure to properly link my libraries before compiling my contract. How do you suggest handling library placeholders with rust-web3?

About my compiler:
```
truffle(develop)> version
Truffle v5.0.2 (core: 5.0.2)
Solidity v0.5.0 (solc-js)
Node v10.15.0
```

elichai:
When you compile linked contracts with truffle it puts the name of the library as ascii into the hex (doesn't make sense to me but ok)

This adds support similar to truffle's migrations `link` function.

Resolves #186.

I'm still testing this with real contracts and real Transport layer but wanted to know what you think.


src/contract/deploy.rs
```diff
@@ -51,12 +54,22 @@ impl<T: Transport> Builder<T> {
         let options = self.options;
         let eth = self.eth;
         let abi = self.abi;
+        let code = code.into();
+
+        let mut code_hex = String::from_utf8(code).unwrap();
```
tomusdrw: Change the signture to be `V: Into<String>` to avoid unsafe conversions (`.unwrap()`)

src/contract/deploy.rs
```diff
@@ -51,12 +54,22 @@ impl<T: Transport> Builder<T> {
...
+
+        for (lib, address) in self.linker {
+            let re = Regex::new(&format!("__{}_+", lib)).unwrap();
```
tomusdrw: Why do we need `Regex` for that? seems that it's a static replacement, no?

elichai: If I understand correctly the amount of underscores is dynamic

src/contract/deploy.rs
elichai: I wanted to support the current execute that accepted a slice.
But I'll change this

tomusdrw: I see, but this `.unwrap()` might be incorrect, I don't think we should do that (it assumes that the `Vec<u8>` is valid UTF8 which might not be the case). I rather think we should change the API.
Either via `Into<String>` or `AsRef<str>` (now I think the latter might be better)

src/contract/deploy.rs
tomusdrw: I think it's rather padded to be 32-bytes long.

elichai: Makes sense, so you prefer that I'll use `len()`, `repeat()` and `replace()` instead of regex?
(and i'll look for some docs saying that's really the padding)

src/contract/deploy.rs
elichai: Ok, will do.

src/contract/deploy.rs
tomusdrw: @elichai either this or just use rust `format!` patterns to fill to desired width: https://doc.rust-lang.org/std/fmt/index.html#fillalignment

elichai:
Did what you asked, 2 things:
1. `str.replacen()` copies the whole string in memory even though we need to replace something the same length. the only way to do this without copying is to use the unsafe `as_mut_vec()` and manually replace the characters.

2. for some reason every change I do makes the response from the `TestTransport` different, is that a problem? I haven't yet tested this with a real ethereum client (although it's a straight forward fix)

src/contract/deploy.rs
```diff
@@ -46,17 +48,26 @@ impl<T: Transport> Builder<T> {
...
+        V: AsRef<str>,
     {
         let options = self.options;
         let eth = self.eth;
         let abi = self.abi;
 
+        let mut code_hex = code.as_ref().to_string();
+
+        for (lib, address) in self.linker {
+            let replace = format!("__{:_<1$}", lib, 38-lib.len());
```
tomusdrw: That might underflow if `lib.len() > 38` I'd rather have an explicit error if that's the case

src/contract/deploy.rs
```diff
@@ -46,17 +48,26 @@ impl<T: Transport> Builder<T> {
...
+            code_hex = code_hex.replacen(&replace, &address.to_hex(), 1);
+        }
+        code_hex = code_hex.replace("\"", "").replace("0x", ""); // This is to fix truffle + serde_json redundant `"` and `0x`
+        let code = code_hex.into_bytes();
```
tomusdrw: That conversion is not valid, we need to convert `str` into `bytes`, but `str` should be hex-encoded, so we need to use `from_hex()` here.

Previously that function accepted raw bytes, so `from_hex()` was happening earlier.

src/contract/deploy.rs
```diff
@@ -46,17 +48,26 @@ impl<T: Transport> Builder<T> {
...
     where
         P: Tokenize,
-        V: Into<Vec<u8>>,
+        V: AsRef<str>,
     {
         let options = self.options;
         let eth = self.eth;
         let abi = self.abi;
 
+        let mut code_hex = code.as_ref().to_string();
```
tomusdrw: we probably should get rid of leading `0x` (if there is any)

src/contract/deploy.rs
```diff
@@ -133,12 +146,12 @@ mod tests {
...
+      "{\"data\":\"0x303130323033303400000000000000000000000000000000000000000000000000000000000f42400000000000000000000000000000000000000000000000000000000000000080000000000000000000000000000000000000000000000000000000000000000300000000000000000000000000000000000000000000000000000000000000c000000000000000000000000000000000000000000000000000000000000000084d7920546f6b656e00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000024d54000000000000000000000000000000000000000000000000000000000000\",\"from\":\"0x0000000000000000000000000000000000000005\",\"value\":\"0x5\"}".into(),
```
tomusdrw: This test should not be changed - currently the change is required due to incorrect `into_bytes()` conversion.

tomusdrw:
And thanks a lot for bearing with me and working on this! :)

src/contract/deploy.rs
elichai: Wow, I don't know how I missed that, you're right, before it wasn't hex and it shouldn't be hex now.

src/contract/deploy.rs
elichai: This is done in the end.
https://github.com/tomusdrw/rust-web3/pull/209/files#diff-0dfcbf3acd5476076104fa446de24e33R63

src/contract/deploy.rs
elichai: Thanks!

elichai:
@tomusdrw Rebased and edited everything you asked for. 
it actually look quite nicer(no unwraps and built in error in ethabi) when it's String and not Vec<u8>

PR #213, opened 2019-04-12
Make `Trace` struct compatible with parity-ethereum rpc

olibearo:
The current struct errors when de serializing traces for `FailedCall`s or `FailedCreate`s because the `result` property isn't optional. This type also isn't capturing the `error` field getting sent by the Ethereum node.

Here is the relevant type/serialization logic on the `parity-ethereum` side:
https://github.com/paritytech/parity-ethereum/blob/master/rpc/src/v1/types/trace.rs#L505-L509


tomusdrw:
Thanks!

PR #229, opened 2019-07-10
Codebase upgrade

vorot93:
This is a very large pull request that includes the upgrade to latest dependency versions and fixes suggested by Clippy.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,19 +13,19 @@ edition = "2018"
 
 [dependencies]
 arrayvec = "0.4.10"
-ethabi = "7.0.0"
-ethereum-types = "0.5.2"
+ethabi = "8.0.0"
+ethereum-types = "0.6.0"
 futures = "0.1.26"
 jsonrpc-core = "11.0.0"
 log = "0.4.6"
-parking_lot = "0.7.1"
+parking_lot = "0.8.0"
 rustc-hex = "2.0.1"
-serde = { version="1.0.90", features = ["derive"] }
+serde = { version = "1.0.90", features = ["derive"] }
 serde_json = "1.0.39"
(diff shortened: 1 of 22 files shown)


Cargo.toml
```diff
@@ -12,34 +12,34 @@ readme = "README.md"
...
-ethabi = "7.0.0"
-ethereum-types = "0.5.2"
-futures = "0.1.26"
-jsonrpc-core = "11.0.0"
-log = "0.4.6"
-parking_lot = "0.7.1"
-rustc-hex = "2.0.1"
-serde = { version="1.0.90", features = ["derive"] }
-serde_json = "1.0.39"
+arrayvec = "0.4"
```
tomusdrw: I'd prefer to keep full versions here. I'm totally fine with having shorter versions in applications/binaries, but not in libraries.
The reason for this is that depending on other libraries cargo might resolve  `0.4` to `0.4.0` and if there are any new features added in say `0.4.1` that we actually use the library won't compile.
So it's better to have a minimal required version encoded here.

vorot93: My concern was the opposite one: explicitly listing minor versions will inevitably cause dependencies to quickly become stale. By leveraging semver we allow the bugfixes in dependencies to automatically roll into dependency graph while keeping breaking changes out.

tomusdrw: Sure, but using `="0.4.10"` syntax only specifies the minimal required version, cargo will resolve to any bugfix release out there (i.e. `0.4.11` is fine in Cargo.lock), so I think we are on the same page.

vorot93: @tomusdrw Seems so. Fixed.

tomusdrw: Perfect :ok_hand:, thanks again!

Issue #238, opened 2019-07-27
Support Infura project secret


steveswing:
Infura Basic auth only requires a project secret as a password and omits a username. The current implementation skips creating basic_auth if username is empty.

PR #239, opened 2019-07-27
#238: Support Infura project secret.

steveswing:
Changed to support Infura project secret that omits a username.

### DIFF
--- a/src/transports/http.rs
+++ b/src/transports/http.rs
@@ -128,14 +128,20 @@ impl Http {
             let url = Url::parse(url)?;
             let user = url.username();
 
-            if !user.is_empty() {
-                let auth = match url.password() {
-                    Some(pass) => format!("{}:{}", user, pass),
-                    None => format!("{}:", user),
-                };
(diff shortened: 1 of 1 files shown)


src/transports/http.rs
```diff
@@ -128,14 +128,20 @@ impl Http {
...
             let user = url.username();
 
-            if !user.is_empty() {
-                let auth = match url.password() {
-                    Some(pass) => format!("{}:{}", user, pass),
-                    None => format!("{}:", user),
-                };
-                Some(HeaderValue::from_str(&format!("Basic {}", base64::encode(&auth)))?)
-            } else {
+            let auth = match url.password() {
```
tomusdrw:
```rust
let auth = format("{}:{}", user, url.password().unwrap_or_default());
if auth == ":" {
  None 
} else {
  Some(...)
}
```

Should work the same and feels less complex to me.

steveswing: reduced complexity as you suggested. Thank you for the feedback and approval. This will make Infura project secret support much easier in our application.

PR #242, opened 2019-07-30
Fail deploy on failed status

### LINKED ISSUE #241: Use `status` field to determine if deployment was successfull
At the moment, contract deployment is considered successful if the `contractAddress` field in the receipt for the deployment transaction is set ([code](https://github.com/tomusdrw/rust-web3/blob/7c80ad3368d9a6d32ea92a91bca963268a16b9c2/src/contract/deploy.rs#L124-L125)). This behavior does not yield correct results with `parity-ethereum`. `parity-ethereum` provides the `contractAddress` even if the contract failed to deploy at least in some circumstances (https://github.com/paritytech/parity-ethereum/issues/10925). However it sets the `status` field in the receipt correctly.

I don’t know if behavior is the same beyond the narrow scenario I tested it with nor do I know if this behavior is intended or consistent with other ethereum implementations.

I would suggest that in addition to checking the `contractAddress` we should also check the status to determine if the deployment was successful. (I’d be willing to make a PR myself)

geigerzaehler:
Deploying a contract is successful only if the `status` field of the receipt is `0x1`. Before, the deployment was considered successful if the contract address was present.

Fixes #241


src/contract/deploy.rs
```diff
@@ -120,9 +121,18 @@ impl<T: Transport> Future for PendingContract<T> {
         let eth = self.eth.take().expect("future polled after ready; qed");
         let abi = self.abi.take().expect("future polled after ready; qed");
 
-        match receipt.contract_address {
-            Some(address) => Ok(Async::Ready(Contract::new(eth, address, abi))),
-            None => Err(Error::ContractDeploymentFailure(receipt.transaction_hash)),
+        let status = receipt.status.ok_or(Error::Api(ApiError::InvalidResponse(String::from(
```
tomusdrw: It's not an error, for chains that don't support `status` the value will be `None`. Please just fall back to `contratAddress` detection in such case as I've suggested in the issue.

PR #251, opened 2019-09-06
Improve error for byte deserialization

geigerzaehler:
We improve the error the error message for deserialization failures of `Byte`.

We drop the length check because `rustc_hex` returns an error when the length is not even.

### DIFF
--- a/src/types/bytes.rs
+++ b/src/types/bytes.rs
@@ -1,5 +1,5 @@
 use rustc_hex::{FromHex, ToHex};
-use serde::de::{Error, Visitor};
+use serde::de::{Error, Unexpected, Visitor};
 use serde::{Deserialize, Deserializer, Serialize, Serializer};
 use std::fmt;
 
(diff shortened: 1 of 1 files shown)


src/types/bytes.rs
```diff
@@ -46,12 +46,11 @@ impl<'a> Visitor<'a> for BytesVisitor {
...
-        if value.len() >= 2 && &value[0..2] == "0x" && value.len() & 1 == 0 {
-            Ok(Bytes(
-                FromHex::from_hex(&value[2..]).map_err(|_| Error::custom("invalid hex"))?,
-            ))
+        if value.len() >= 2 && &value[0..2] == "0x" {
+            let bytes = FromHex::from_hex(&value[2..]).map_err(|e| Error::custom(format!("Invalid hex: {}", e)))?;
+            Ok(Bytes(bytes))
         } else {
-            Err(Error::custom("invalid format"))
+            Err(Error::invalid_value(Unexpected::Str(value), &"0x prefix"))
```
tomusdrw: In principle this will also be displayed if the length is less than 2, but it's strictly better than what we have currently.

tomusdrw:
Thanks!

PR #252, opened 2019-09-08
Add `web3::contract::deploy::execute_no_unlock`

rockbmb:
This PR introduces a variant of `web3::contract::deploy::execute` that does not require the account for which it is used to be unlocked.

### DIFF
--- a/src/contract/deploy.rs
+++ b/src/contract/deploy.rs
@@ -44,11 +44,12 @@ impl<T: Transport> Builder<T> {
         self
     }
 
-    /// Execute deployment passing code and contructor parameters.
-    pub fn execute<P, V>(self, code: V, params: P, from: Address) -> Result<PendingContract<T>, ethabi::Error>
+    fn execute_core<P, V, F>(self, code: V, params: P, from: Address, f: F) -> Result<PendingContract<T>, ethabi::Error>
     where
         P: Tokenize,
         V: AsRef<str>,
+        T: Transport,
+        F: Fn(TransactionRequest, T, core::time::Duration, usize) -> confirm::SendTransactionWithConfirmation<T>,
     {
         let options = self.options;
         let eth = self.eth;
@@ -89,11 +90,11 @@ impl<T: Transport> Builder<T> {
             condition: options.condition,
         };
 
-        let waiting = confirm::send_transaction_with_confirmation(
-            eth.transport().clone(),
+        let waiting = f(
             tx,
-            self.poll_interval,
-            self.confirmations,
+            eth.transport().clone(),
+            self.poll_interval.clone(),
+            self.confirmations.clone(),
         );
 
         Ok(PendingContract {
@@ -101,6 +102,19 @@ impl<T: Transport> Builder<T> {
             abi: Some(abi),
             waiting,
         })
+    }
+
+    /// Execute deployment passing code and contructor parameters.
+    pub fn execute<P, V>(self, code: V, params: P, from: Address) -> Result<PendingContract<T>, ethabi::Error>
+    where
+        P: Tokenize,
+        V: AsRef<str>,
+        T: Transport,
+    {
+        let f = |tx, eth: T, poll_interval, confirmations| {
+            confirm::send_transaction_with_confirmation(eth, tx, poll_interval, confirmations)
+        };
+        self.execute_core(code, params, from, f)
     }
 
     /// Execute deployment passing code and contructor parameters.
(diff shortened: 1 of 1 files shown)


src/contract/deploy.rs
```diff
@@ -102,6 +103,50 @@ impl<T: Transport> Builder<T> {
...
+    where
+        P: Tokenize,
+        V: AsRef<str>,
+        T: Transport,
+    {
+        let f = |tx, eth: T, poll_interval, confirmations| {
+            let raw_tx = web3
+                .personal()
+                .sign_transaction(tx, password)
+                .wait()
```
tomusdrw: That is making a synchronous call, it's not acceptable in async context.

src/contract/deploy.rs
```diff
@@ -102,6 +103,50 @@ impl<T: Transport> Builder<T> {
...
+        self.execute_core(code, params, from, f)
+    }
+
+    /// Execute deployment passing code and contructor parameters.
+    ///
+    /// Unlike the above `execute`, this method uses
+    /// `sign_raw_transaction_with_confirmation` instead of
+    /// `sign_transaction_with_confirmation`, which requires the account from
+    /// which the transaction is sent to be unlocked.
+    pub fn execute_no_unlock<P, V>(
```
tomusdrw: Can we call it `sign_and_execute`?

src/contract/deploy.rs
```diff
@@ -102,6 +103,50 @@ impl<T: Transport> Builder<T> {
...
+        P: Tokenize,
+        V: AsRef<str>,
+        T: Transport,
+    {
+        let f = |tx, eth: T, poll_interval, confirmations| {
+            let raw_tx = web3
+                .personal()
+                .sign_transaction(tx, password)
+                .wait()
+                .expect("Transaction signing failed");
```
tomusdrw: We definitely should not panic if we can't sign the transaction - it should just be a regular `Err` that can be later handled by the caller.

src/contract/deploy.rs
rockbmb: Sure!

src/contract/deploy.rs
rockbmb: You are right.

src/contract/deploy.rs
rockbmb: Then how should I fetch the value that `web3.personal().sign_transaction()` returns?

I can chain this computation with `and_then(move |signed_tx| {send_raw_transaction_with_confirmation ... }`, but that's only moving the `Future` down the line, and the closure can't return one.

rockbmb: Almost forgot: the `impl<T: Transport> Future` for `SendTransactionWithConfirmation<T>` has `type Item = TransactionReceipt`, meaning even if the above is done, we will still be getting a `TransactionReceipt` where we should be getting a `SendTransactionWithConfirmation<T>`.

tomusdrw: Most likely we need to change `PendingContract` to be able to work with aribtrary `Future` that resolves to `TransactionReceipt`:
```
PendingContract<
  T: Transport, 
  F: Future<Item=TransactionReceipt, Error=...> = SendTransactionWithConfirmation<T>,
> {

}
```
and then use the `and_then`combinator as you've mentioned.

rockbmb: I was reluctant to change `PendingContract` without the go-ahead of someone more knowledgeable about the code base than me. If it sounds reasonable, then I will do so, and re-request a review.

Thank you for your attention :+1:

rockbmb: I understand that the type signatures you wrote here are merely suggestive, but what does `F: Future<Item=TransactionReceipt, Error=...> = SendTransactionWithConfirmation<T>` mean? I have not come across this syntax before.

tomusdrw: It's a generic parameter, but with a default value. So that `PendingContract<T>` would still be backward compatible, as the `F` type would default to `SendTransactionWithConfirmation` if omitted.

rockbmb: I split the two functions back again since the previous refactoring would not work with this change to `PendingContract`.

However, now I'm getting an error on [this line](https://github.com/tomusdrw/rust-web3/pull/252/files#diff-0dfcbf3acd5476076104fa446de24e33R177) I am not sure I understand:

```rust
F
mismatched types

expected type parameter, found struct `futures::AndThen`

note: expected type `Ft`
         found type `futures::AndThen<helpers::CallFuture<types::transaction::RawTransaction, <T as Transport>::Out>, confirm::SendTransactionWithConfirmation<T>, [closure@src\contract\deploy.rs:165:79: 172:10 eth:_, self:_]>`rustc(E0308)
deploy.rs(177, 13): expected type parameter, found struct `futures::AndThen`
```

I was under the impression `and_then` returned a struct, `AndThen`, that implemented `Future` so I am not sure why `PendingContract` would have issues with it.

Do you mind lending me a hand with this? I apologise for my unreasonable request.

tomusdrw:
@rockbmb good to go now?

rockbmb:
@tomusdrw not quite. I'm getting an odd compiler error at `sign_and_execute`'s callsite, and I can't quite figure it out.

tomusdrw:
Maybe add an example/test to this PR so that we can debug it together?

rockbmb:
@tomusdrw I'll add one right next to the test for `execute`, but sadly this will have to wait - I'll be afk for a few hours.

src/contract/deploy.rs
```diff
@@ -49,6 +49,61 @@ impl<T: Transport> Builder<T> {
...
+            confirm::send_transaction_with_confirmation(transport, tx, poll_interval, confirmations)
+        })
+    }
+    /// Execute deployment passing code and contructor parameters.
+    ///
+    /// Unlike the above `execute`, this method uses
+    /// `sign_raw_transaction_with_confirmation` instead of
+    /// `sign_transaction_with_confirmation`, which requires the account from
+    /// which the transaction is sent to be unlocked.
+    pub fn sign_and_execute<P, V, Ft>(
```
geigerzaehler: The `Ft` parameter is unused. This is problematic for calling code because the compiler cannot infer the instantiation of `Ft`. This leads to the calling code always requiring an explicit type annotation.

tomusdrw: Let's just remove it then. I guess I left it by mistake.

rockbmb: I'll try this out.

tomusdrw: Does it solve the issues you were having?

rockbmb: It did. I get other errors now, but as far as this PR goes, it has what we need.
Do you still want me to add a test before we merge?

tomusdrw:
@rockbmb Feel free to add a test in a follow-up PR.

Issue #114, opened 2018-04-18
make block number U256 everywhere


snd:
block number is `U256` everywhere (https://tomusdrw.github.io/rust-web3/web3/types/enum.BlockNumber.html?search=block_number) except for `BlockNumber::Number(u64)` (https://tomusdrw.github.io/rust-web3/web3/types/enum.BlockNumber.html) where it's a `u64`. i think it would be good to have this consistent. if this is a change that's wanted i'll open a PR.

tomusdrw:
Yeah, I think the spec say it should always be `Quantity` (not 100% sure about input parameters though), and we should probably use `U64` since it's more than enough for the block number (probably we haven't used it in the past due to lack of this type).

snd:
by `U64` you mean the `U64` in ethereum-types, right?
what's the advantage this type has over `u64`?

right, `64` bits is big enough for the block number

tomusdrw:
`U64` is JSON (de)serializable to hex and not just number.

tomusdrw:
Done, we switched to `U128`.

PR #253, opened 2019-09-08
Updated blocks numbers to use U64

azastrael:
Hey :hand:!

The following methods/types/fields were using `U256` to represent a block's number instead of, as everywhere else and decided [here](https://github.com/tomusdrw/rust-web3/issues/114#issuecomment-510003843), `U128`:
- `web3::api::Eth::block_number`
- `web3::api::Web3::wait_for_confirmations`
- `web3::confirm::ConfirmationCheck::Check` and all related structs/types
- `web3::types::TransactionReceipt.block_number`

Also, maybe `web3::types::BlockNumber` should use `U128` instead of `u64`?

### DIFF
--- a/src/api/eth.rs
+++ b/src/api/eth.rs
@@ -4,7 +4,7 @@ use crate::api::Namespace;
 use crate::helpers::{self, CallFuture};
 use crate::types::{
     Address, Block, BlockId, BlockNumber, Bytes, CallRequest, Filter, Index, Log, SyncState, Transaction,
-    TransactionId, TransactionReceipt, TransactionRequest, Work, H256, H520, H64, U256,
+    TransactionId, TransactionReceipt, TransactionRequest, Work, H256, H520, H64, U128, U256,
 };
 use crate::Transport;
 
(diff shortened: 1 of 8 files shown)


tomusdrw:
The PR looks good, but can we use `U64` instead? The use of `U256/U128` is a bit historical (we didn't use the `ethereum-types` crate in the past, so the types had to be manually implemented).

azastrael:
@tomusdrw 

> The PR looks good, but can we use `U64` instead? The use of `U256/U128` is a bit historical (we didn't use the `ethereum-types` crate in the past, so the types had to be manually implemented).

Sure :smiley:. Also, should we also make `web3::types::BlockNumber`'s `Number(u64)` variant use `U64` instead of `u64`(it would make more sense imho)?

tomusdrw:
> Sure smiley. Also, should we also make web3::types::BlockNumber's Number(u64) variant use U64 instead of u64(it would make more sense imho)?

Yes, let's do that.

azastrael:
@tomusdrw 

> Yes, let's do that.

There's also `types::Trace.block_number` but making it use `U64` would need us to implement `Deserialize` on `types::Trace` because Parity encodes it as a base10 number (instead of a base16 one).

tomusdrw:
> @tomusdrw
> 
> > Yes, let's do that.
> 
> There's also `types::Trace.block_number` but making it use `U64` would need us to implement `Deserialize` on `types::Trace` because Parity encodes it as a base10 number (instead of a base16 one).

I see. I think that one can be left as-is.

PR #257, opened 2019-09-17
Change future instance for SendTransactionWithConfirmation

### LINKED ISSUE #256: send_transaction_with_confirmation with 0 confirmations for development
I’d like to use `send_transaction_with_confirmation` with Parity and a [private development chain](https://wiki.parity.io/Private-development-chain.html) (`parity --config dev`). This development chain mines a block immediately after a transaction arrives and it only mines blocks when transaction arrives. This means that I cannot use `send_transaction_with_confirmation` with a confirmation count of 1. Since only new transaction trigger the creation of a block the future returned from the call will only resolve once a another transaction is sent. This is insufficient in, say, a test scenario, where the last transaction will never be confirmed.

The alternative is to use `send_transaction_with_confirmation` with a confirmation count of 0. However, this call will not wait for the transaction to be included in a block. It will ask for the receipt immediately after the transaction is sent. If the block has not been mined yet then there will be no receipt and the call panics with `receipt can't be null after wait for confirmations; qed`. (This error occurs in fact with other chain configuration, including the ethereum mainnet. However using zero confirmations with ethereum has limited use.)

### Solution

To fix the problem we would need to ensure that for a zero confirmation count the receipt is retrieved only when a block including the transaction has been mined. This would require a change to `SendTransactionWithConfirmation::poll`.

One solution could be to repeatedly poll for the receipt and return only when it is available. Alternatively we could call `wait_for_confirmations` _before_ the transaction is sent to ensure that we receive an update when the block containing the transaction is mined.

//cc @rockbmb

rockbmb:
Closes #256.

### DIFF
--- a/src/confirm.rs
+++ b/src/confirm.rs
@@ -7,7 +7,7 @@ use crate::helpers::CallFuture;
 use crate::types::{Bytes, TransactionReceipt, TransactionRequest, H256, U64};
 use crate::{Error, Transport};
 use futures::stream::Skip;
-use futures::{Future, IntoFuture, Poll, Stream};
+use futures::{Async, Future, IntoFuture, Poll, Stream};
 
 /// Checks whether an event has been confirmed.
 pub trait ConfirmationCheck {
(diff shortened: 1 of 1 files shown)


src/confirm.rs
```diff
@@ -288,9 +288,17 @@ impl<T: Transport> Future for SendTransactionWithConfirmation<T> {
...
                 }
                 SendTransactionWithConfirmationState::GetTransactionReceipt(ref mut future) => {
-                    let receipt = try_ready!(Future::poll(future))
-                        .expect("receipt can't be null after wait for confirmations; qed");
-                    return Ok(receipt.into());
+                    let err = "receipt can't be null after wait for confirmations; qed";
+                    if self.confirmations == 0 {
+                        match Future::poll(future)? {
+                            Async::Ready(opt_receipt) => return Ok(opt_receipt.expect(err).into()),
+                            Async::NotReady => SendTransactionWithConfirmationState::GetTransactionReceipt(*future),
```
rockbmb: Using `ref mut future` is unavoidable in the arguments of the match arms, but dereferencing `*future` fails because there is no `Copy` for its type, and it is not feasible to do so.

Any suggestions appreciated.

geigerzaehler: I don’t know how to satisfy the borrow checker here either. However, I doubt that even if this would compile this accomplishes what you want. With this implementation the code will effectively loop and block until a receipt is received. First this is counter to how futures work, and second this will not re-fetch a receipt and solve the issue.

tomusdrw:
Closing as stale.

PR #258, opened 2019-09-17
Add test to sign_and_execute

rockbmb:
@tomusdrw still WIP.

Is there anything I need to enable or do in order to use account creation with the test transport?

### DIFF
--- a/src/contract/deploy.rs
+++ b/src/contract/deploy.rs
@@ -260,6 +260,75 @@ mod tests {
     }
 
     #[test]
+    fn should_sign_and_deploy_a_contract() {
+        // given
+        let mut transport = TestTransport::default();
+        // Transaction Hash
+        transport.add_response(rpc::Value::String(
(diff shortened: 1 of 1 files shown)


src/contract/deploy.rs
```diff
@@ -259,6 +259,75 @@ mod tests {
...
+        // block number
+        transport.add_response(rpc::Value::String("0x25a".into()));
+        // receipt again
+        transport.add_response(receipt);
+
+        {
+
+            let password = "";
+            let web3 = crate::api::Personal::new(transport.clone());
+            let addr = web3.new_account(password).wait().unwrap();
```
tomusdrw: What's the purpose of this? Test transport has all the responses mocked, so that will just return one of the responses you've added earlier. Use some hardcoded address here instead of calling anything.

tomusdrw:
Closing as stale.

PR #260, opened 2019-09-21
Updated a few trace-related types to make their fields public

azastrael:
Hey :wave:!

I updated a few trace-related types to make their fields public :smiley:.


azastrael:
Also, I was wondering whether we should make the fields of `Filter` and `TraceFilter` public.

PR #261, opened 2019-09-21
Removed Res::FailedCallOrCreate and added missing fields to TransactionTrace

azastrael:
Hey :wave:!

Parity encodes traces results in one of two fields:
- "result" which can either store a `CallResult`, `CreateResult` or be equal to `null` (`Res::None`)
- "error" which stores a `TraceError` (https://github.com/paritytech/parity-ethereum/blob/master/ethcore/trace/src/types/error.rs#L25) and which Parity considers as either a `FailedCall` or `FailedCreate`

Currently, `Res` would only be able to decode a `CallResult`, `CreateResult` or `None` because it would only look into the "result" field. Thus, `FailedCallOrCreate` is unnecessary but `Trace` and `TransactionTrace` need an `error` field (`TransactionTrace` didn't have one).

---

I also added an `action_type` field to `TransactionTrace`.

### DIFF
--- a/src/types/trace_filtering.rs
+++ b/src/types/trace_filtering.rs
@@ -114,8 +114,6 @@ pub enum Res {
     Call(CallResult),
     /// Create
     Create(CreateResult),
-    /// Call or Create failure
-    FailedCallOrCreate(String),
     /// None
     None,
 }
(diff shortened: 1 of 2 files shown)


azastrael:
Maybe making our `Res` have the same variants as Parity's (https://github.com/paritytech/parity-ethereum/blob/master/rpc/src/v1/types/trace.rs#L434) could be done but it would probably need us to implement `Deserialize` for both `Trace` and `TraceFiltering` (except if there's a way to get the current JSON field's key when deserializing `Res`).

We also could update the `error` fields from a `String` to a `TraceError` (https://github.com/paritytech/parity-ethereum/blob/master/ethcore/trace/src/types/error.rs#L25).

Issue #203, opened 2019-03-06
trace_replayBlockTransactions does not work


olibearo:
The method name was missing an s, it was `trace_replayBlockTransaction`.

After fixing the method name, the response had a deserialization error.

__rust-web3 error__:

```
Error(Decoder("Error(\"invalid type: map, expected a 0x-prefixed hex-encoded vector of bytes\", line: 0, column: 0)"), State { next_error: None, backtrace: InternalBacktrace { backtrace: None } })
```

__Relevant parity-ethereum logs__

```
2019-03-06 14:37:14 UTC jsonrpc-eventloop-1 TRACE rpc  Request: {"jsonrpc":"2.0","method":"trace_replayBlockTransactions","params":["latest",["trace"]],"id":0}
```

```
[...]

PR #262, opened 2019-09-21
Fixed the "replay_block_transactions" method of api::Traces and added missing fields to BlockTrace

azastrael:
Hey :wave:!

The method was calling "trace_replayBlockTransaction" instead of "trace_replayBlockTransactions" (missing "s" at the end) and was returning a single BlockTrace instead of a vector of them (https://wiki.parity.io/JSONRPC-trace-module.html#trace_replayblocktransactions).

Tests were also added for both the method (using the block 46147) and deserialization of vectors of BlockTrace.

It fixes the RPC API part of #203 and I tried decoding the output that the OP gave and apart from a `,"id":0}` at the end that shouldn't have been there, everything was fine.

NOTE: the block I chose isn't the best possible so feel free to suggest change it (beware of the size of the JSON though).

---

I also added a `transaction_hash` field to `BlockTrace` which is returned by Parity when calling "trace_replayBlockTransactions".

### DIFF
--- a/src/api/traces.rs
+++ b/src/api/traces.rs
@@ -135,17 +135,17 @@ mod tests {
 
     const EXAMPLE_BLOCKTRACES: &'static str = r#"
 	[{
-        "output": "0x010203",
+        "output": "0x",
         "stateDiff": null,
         "trace": [
             {
                 "action": {
                     "callType": "call",
-                    "from": "0x0000000000000000000000000000000000000000",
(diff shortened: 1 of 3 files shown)


azastrael:
Also, `src/types/example-trace-str.rs` and `src/types/example-traces-str.rs` should probably be moved to another directory (it pollutes this one IMO)

PR #273, opened 2019-10-15
Implemented From<H256> for TransactionId

azastrael:
Hey :wave:!

I just implemented `From<H256>` for `TransactionId` which returns `TransactionId::Hash(_)` :smiley:.


src/types/transaction_id.rs
```diff
@@ -8,3 +8,9 @@ pub enum TransactionId {
     /// By block and index
     Block(BlockId, Index),
 }
+
+impl From<H256> for TransactionId {
+    fn from(hash: H256) -> Self {
+        TransactionId::Hash(hash)
+    }
+}
```
tomusdrw:
```suggestion
}

```

PR #277, opened 2019-10-29
added `Tokenizable` implementation to most primitive int and uint types

nlordell:
Added implementation of `Tokenizable` trait to most primitive (u)int rust types as they are also valid solidity types. There is an issue with implementing it for `u8` - see note in code. The conflicting trait implementations can be worked around by introducing a second tokenizable trait for types that can be part of `(Fixed)Array` tokens (so everything except u8).

### DIFF
--- a/src/contract/tokens.rs
+++ b/src/contract/tokens.rs
@@ -191,7 +191,7 @@ impl Tokenizable for Address {
     }
 }
 
-macro_rules! uint_tokenizable {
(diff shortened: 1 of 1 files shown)


nlordell:
Not sure why the build is failing on beta and nightly - it is complaining that the build stalled and no output was received for the last 10 minutes. Seems to build on stable though :).

PR #279, opened 2019-11-06
Add transaction signing with in `accounts` sub-namespace.

### LINKED ISSUE #116: Signing Transactions Offline

nlordell:
Also adds a few related methods like `hash_message`, `sign` and `recover`. This closes  #116.


reuvenpo:
Hey there @nlordell, I couldn't find a way to contact you so I'm writing this here.

My team is trying to achieve this exact functionality this week so that we can use our own private key instead of relying on unlocked accounts on ethereum nodes.
We have a very ugly wrapper code which uses `ethereum-tx-sign` plus parts of this crate (`web3`) that we're working on. I figured we couldn't be the first team to need this functionality, so i decided to check this repo's pull request list, and was really happy to see this work is already being done :smile: 

We want to ask when can we expect this to be merged into master and published in a new release?

We also have a question that may be better asked of @tomusdrw: what is the best way, given a contract's abi, to sign calls to the contract's methods? ideally, we would like to have a `signed_call_with_confirmations` similar to `web3::contract::Contract::call_with_confirmations` which would use the functionality introduced in this PR.

tomusdrw:
Hi guys! I didn't have time to look closely at the PR but will definitely do that in the upcoming 36h.

src/api/accounts.rs
```diff
@@ -0,0 +1,515 @@
...
+    to: Address,
+    /// Supplied gas (None for sensible default)
+    gas: U256,
+    /// Gas price (None for sensible default)
+    gas_price: U256,
+    /// Transfered value (None for no transfer)
+    value: U256,
+    /// Data (None for empty data)
+    data: Bytes,
+}
```
tomusdrw: I've recently worked on a similar set of primitives:
https://github.com/tomusdrw/jsonrpc-proxy/blob/master/ethereum-proxy/plugins/accounts/transaction/src/lib.rs

Just published that to crates.io as `ethereum-transaction`. Feel free to re-use.

src/api/accounts.rs
```diff
@@ -0,0 +1,515 @@
...
+
+        assert_eq!(signed.into_signature(), expected_signature);
+        assert_eq!((v, r, s).into_signature(), expected_signature);
+        assert_eq!(signed.signature.into_signature(), expected_signature);
+    }
+
+    #[test]
+    fn raw_tx_params_sign() {
+        // retrieved test vector from:
+        // https://web3js.readthedocs.io/en/v1.2.2/web3-eth-accounts.html#eth-accounts-signtransaction
```
tomusdrw: :heart: Thank you for all the tests.

src/api/mod.rs
```diff
@@ -1,5 +1,6 @@
 //! `Web3` implementation
 
+mod accounts;
```
tomusdrw: Please re-export `accounts::Accounts` otherwise the docs are not visible.

src/error.rs
```diff
@@ -25,6 +26,9 @@ pub enum Error {
     /// io error
     #[display(fmt = "IO error: {}", _0)]
     Io(IoError),
+    /// signing error
+    #[display(fmt = "Signing error: {}", _0)]
+    Signing(String),
```
tomusdrw: Why not expose `EthsignError` instead?

src/types/signed.rs
```diff
@@ -0,0 +1,55 @@
...
+pub struct TransactionParameters {
+    /// Transaction nonce (None for account transaction count)
+    pub nonce: Option<U256>,
+    /// To address
+    pub to: Option<Address>,
+    /// Supplied gas (None for sensible default)
+    pub gas: Option<U256>,
+    /// Gas price (None for sensible default)
+    pub gas_price: Option<U256>,
+    /// Transfered value (None for no transfer)
```
tomusdrw: Why `None` and not `H256::zero()` or `Default::default()`? I think it makes it really confusing if there is any difference between `None` and `Some(0)` (afaik there isn't).

src/types/signed.rs
```diff
@@ -0,0 +1,55 @@
...
+    pub value: Option<U256>,
+    /// Data (None for empty data)
```
tomusdrw: Same here. I think it would be best to drop the `Option`.

src/api/accounts.rs
```diff
@@ -0,0 +1,515 @@
...
+            message_hash: message_hash.into(),
+            v,
+            r: signature.r.into(),
+            s: signature.s.into(),
+            signature: signature_bytes,
+        })
+    }
+
+    /// Recovers the Ethereum address which was used to sign the given data.
+    pub fn recover<S, Sig>(&self, message: S, signature: Sig) -> Result<Address, Error>
```
tomusdrw: Might be useful to have another `recover` for raw hashes? Maybe make the `message` an `enum` and allow passing etiher `Hash` or `AsRef<str>`?

tomusdrw:
> ideally, we would like to have a signed_call_with_confirmations method similar to web3::contract::Contract::call_with_confirmations which would use the functionality introduced in this PR.

Happy to accept a PR introducing this feature @reuvenpo.

src/api/accounts.rs
```diff
@@ -0,0 +1,515 @@
...
+    where
+        S: AsRef<str>,
+    {
+        let message = message.as_ref().to_string();
+        let message_hash = self.hash_message(&message);
+
+        let signature = key.sign(&message_hash[..]).map_err(EthsignError::from)?;
+        // this is what web3.js does ¯\_(ツ)_/¯, it is documented here:
+        // https://web3js.readthedocs.io/en/v1.2.2/web3-eth-accounts.html#sign
+        let v = signature.v + 27;
```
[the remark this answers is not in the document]
nlordell: Thanks for the explanation! TIL.

src/api/accounts.rs
```diff
@@ -0,0 +1,515 @@
...
+}
+
+impl<T: Transport> Accounts<T> {
+    /// Gets the parent `web3` namespace
+    fn web3(&self) -> Web3<T> {
+        Web3::new(self.transport.clone())
+    }
+
+    /// Signs an Ethereum transaction with a given private key.
+    pub fn sign_transaction(&self, tx: TransactionParameters, key: &SecretKey) -> SignTransactionFuture<T> {
```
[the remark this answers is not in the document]
nlordell: I chose to implement a future type here since I didn't see `impl Future` used anywhere else. Also I don't think the use of `async`/`await` would be good here since this crate uses futures 0.1 which require a compat layer to work with `async`/`await`.

src/api/mod.rs
nlordell: Oops!

src/api/accounts.rs
```diff
@@ -0,0 +1,515 @@
...
+    fn poll(&mut self) -> Poll<Self::Item, Self::Error> {
+        let (nonce, gas_price, chain_id) = try_ready!(self.inner.poll());
+
+        let chain_id = chain_id.parse::<u64>().map_err(|e| Error::Decoder(e.to_string()))?;
+        let tx = RawTxParameters {
+            nonce,
+            to: self.tx.to.unwrap_or_default(),
+            value: self.tx.value.unwrap_or_default(),
+            gas_price,
+            gas: self.tx.gas.unwrap_or(TRANSACTION_BASE_GAS_FEE),
```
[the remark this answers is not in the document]
nlordell: I chose 21K as this is the base transaction amount and exactly what it uses to send eth from one address to another (with no contracts involved). I think if we want to be extra cleaver, we could estimate the gas price when it is not provided.

src/error.rs
nlordell: I tried but ran into some issues. `EthsignError` doesn't implement `Clone` or `PartialEq`. Let me take a look, maybe I can come up with something.

src/types/signed.rs
nlordell: For gas price specifically there is a difference. `Some(0)` means that you want to pay `0` for gas (transaction will never be mined though). `None` on the other hand will tell the signing function to get the current recommended gas price and use that.

src/types/signed.rs
nlordell: Sounds good. We can drop `Option` for `to`, `value`, and `data`. I had used `Option` originally to signal that it providing this transaction parameter was optional and when it was not provided some default would be used. For the aforementioned fields, the default also happens to be `Default::default()` so we can get rid of `Option` for them.

src/api/accounts.rs
nlordell: Oops! That was a bit of an oversight.

src/api/accounts.rs
```diff
@@ -0,0 +1,515 @@
...
+        Ok(public_key.address().into())
+    }
+}
+
+type MaybeReady<T, R> = Either<FutureResult<R, Error>, CallFuture<R, <T as Transport>::Out>>;
+
+type TxParams<T> = Join3<MaybeReady<T, U256>, MaybeReady<T, U256>, MaybeReady<T, String>>;
+
+/// Future resolving when transaction signing is complete
+pub struct SignTransactionFuture<T: Transport> {
```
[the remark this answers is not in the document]
nlordell: I can change this to `impl Trait` if you feel strongly about it, but I don't think `async`/`await` is a good idea here because that would mean the crate would mix futures 0.1 and std futures depending on what is being called.

src/api/accounts.rs
```diff
@@ -0,0 +1,515 @@
...
+
+        let signed = tx.sign(&self.key, Some(chain_id))?;
+
+        Ok(Async::Ready(signed))
+    }
+}
+
+/// Trait for converting data into an `ethsign::Signature`.
+pub trait IntoSignature {
+    fn into_signature(self) -> Signature;
```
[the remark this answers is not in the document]
nlordell: Oops!

src/api/accounts.rs
```diff
@@ -0,0 +1,515 @@
...
+            r: r.into(),
+            s: s.into(),
+        }
+    }
+}
+
+impl<'a> IntoSignature for &'a Bytes {
+    fn into_signature(self) -> Signature {
+        if self.0.len() != 65 {
+            panic!("invalid signature bytes");
```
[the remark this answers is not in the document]
nlordell: Alright, that sounds like a good idea.

src/api/accounts.rs
nlordell: This is great! I will change to use this instead.

src/api/accounts.rs
```diff
@@ -0,0 +1,515 @@
...
+        (self.v, self.r, self.s).into_signature()
+    }
+}
+
+/// Raw transaction data for signing. When a transaction is signed, all
+/// parameter values need to be finalized for signing, and all parameters are
+/// required. Note that transaction data does not actually have the `from`
+/// public address as that is recoverable from the signature.
+#[derive(Debug)]
+struct RawTxParameters {
```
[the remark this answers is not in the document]
nlordell: Oops!

src/api/accounts.rs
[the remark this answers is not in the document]
reuvenpo: I think it's appropriate to port to the async/.await notation as a separate task, as that would ideally just rewrite 90% of the async stuff

reuvenpo:
> > ideally, we would like to have a signed_call_with_confirmations method similar to web3::contract::Contract::call_with_confirmations which would use the functionality introduced in this PR.
> 
> Happy to accept a PR introducing this feature @reuvenpo.

I'll build it on this PR, so i'll wait until this one has finished cooking :)

src/types/signed.rs
tomusdrw: I'd keep it for `to` as well. There is a difference between contract creation and sending to `0x0` address.

src/types/signed.rs
tomusdrw: Yes, I only meant `value` and `data` field. The rest is fine.

src/types/signed.rs
nlordell: Ah I see it in `ethereum_transaction`, it gets encoded differently if it's there or if its missing.

src/api/accounts.rs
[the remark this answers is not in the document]
tomusdrw: I'd leave the estimation to the user - we can suggest them to run `estimateGas` before making a call. BTW is it easy to go from the payload required for `estimateGas` and the payload that goes to this function? We should consider adding a `From` implementations to make such workflow easier.

Indeed `21k` is ok for simple transfers, but if only you'll try to send some data, we will need more gas. I'm still in favour of increasing the "default sensible" gas here - the user will get the rest refunded anyway and I think it can prevent a class of simple errors. Alternatively we can make `gas` a required field and force the user to always set it.

src/api/accounts.rs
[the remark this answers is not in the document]
tomusdrw: Not feeling strong, the codebase is not migrated indeed. I'm fine with the futures as it is currently.

src/api/accounts.rs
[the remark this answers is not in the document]
tomusdrw: Makes sense!

src/error.rs
tomusdrw: I see. I don't really remember why we need both `Clone` and `PartialEq` here. If it's possible let's remove that if not, then could you file a separate issue for this, so that we can later change `String` to proper error?

src/api/accounts.rs
nlordell: I just noticed that `ethereum-transaction` is using a different `ethereum-types` version which causes some issues. Are the plans on bumping `ethereum-types` version for `ethereum-transaction`?

tomusdrw: https://github.com/tomusdrw/jsonrpc-proxy/pull/93 Publishing as we speak.

src/api/accounts.rs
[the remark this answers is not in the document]
nlordell: FWIW, web3js requires gas parameter for `web3.eth.accounts.signTransaction`. I like having a default gas as I find it makes things more ergonomic (so we can implement `Default` for `TransactionParameters`). I think I'll hike it up to `100k` then.

I'll add the `Into` and `From` impls between `TransactionParameters` and `CallRequest` which is used for estimate gas.

src/error.rs
nlordell: OK. That sounds like a good plan. There are other errors that are also converted to string that might benefit from having concrete error types as well (like `Error::Decoder`, `Error::Transport`).

reuvenpo: I don't know what the policy on defining error types in this crate is, but i can see that this crate doesn't use any special magic for its errors. Using something like `error-chain`+`error-chain-derive`, `derive-error`, `failure`, the new and shiny `anyhow` could make error definition and handling smoother.
(I think the ones i mention before `failure` are unmaintained)

tomusdrw: We've been actually getting rid of all the crates you've mentioned from other projects as they don't bring that much value. I'm a fan of a simple enum + `derive_more` to reduce boilerplate now.

src/api/accounts.rs
nlordell: Awesome-sauce

src/api/accounts.rs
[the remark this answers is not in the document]
nlordell: @tomusdrw Actually, thinking about it now it would be nice to be able to do something like:
```rust
let signed_data = accounts.sign("hello world", &key);
let address = accounts.recover(signed_data)?;
assert_eq!(address, key.public().address());
let signed_tx = accounts.sign(tx, &key).await?;
let address = accounts.recover(signed_tx)?;
assert_eq!(address, key.public().address());
```

What do you think?

src/error.rs
reuvenpo: I think the biggest strength of those crates is the fact they provide backtraces for chains of errors. But of course i won't insist on this.

tomusdrw: Yeah, that's what you get if you implement `std::error::Error` and provide `source`. I agree that it's nice when automated, but as I said, I'm currently more inclined to just implement it manually given my bad experiences with both `error-chain` and `failure`.

nlordell: A little off-topic but I really like `thiserror`

nlordell: A little off-topic but I really like `thiserror`

src/api/accounts.rs
nlordell: @tomusdrw Actually, thinking about it now it would be nice to be able to do something like:
```rust
let signed_data = accounts.sign("hello world", &key);
let address = accounts.recover(signed_data)?;
assert_eq!(address, key.public().address());
let signed_tx = accounts.sign(tx, &key).await?;
let address = accounts.recover(signed_tx)?;
assert_eq!(address, key.public().address());
```

What do you think?

tomusdrw: Yes, but how do you recover something you didn't sign yourself? Also since we don't have specialisation either the method names need to be different or you need to accept `impl Into<SomeEnum>` as a parameter.

nlordell: I went ahead and implemented this. If you don't like it, let me know and I can change it.

nlordell: Sorry, for some reason I didn't see you message when I commented. Yes you can recover things that you did not sign yourself. A `Recovery` struct with the recovery data can be constructed fairly easily, we do this in one of the unit tests.

Let me know if you like how I implemented it, if not I can change it to something else.

nlordell:
@tomusdrw I think I addressed all of your comments. Note that I changed how recovery works, let me know what you think of the new implementation. Take a look at the `accounts_recovery*` tests to see how it works :smile:

src/api/accounts.rs
```diff
@@ -0,0 +1,466 @@
...
+    pub fn sign_transaction(&self, tx: TransactionParameters, key: &SecretKey) -> SignTransactionFuture<T> {
+        SignTransactionFuture::new(self, tx, key)
+    }
+
+    /// Hash a message according to EIP-191.
+    ///
+    /// The data is a UTF-8 encoded string and will enveloped as follows:
+    /// `"\x19Ethereum Signed Message:\n" + message.length + message` and hashed
+    /// using keccak256.
+    pub fn hash_message<S>(&self, message: S) -> H256
```
tomusdrw: We can now use `impl Trait` for this:
```suggestion
    pub fn hash_message(&self, message: impl AsRef<str>) -> H256 {
```

Most likely it applies to every generic method introduced in this PR. But it's not critical, just a hint :)

src/error.rs
tomusdrw: @nlordell Thanks for the hint, indeed looks very nice.

src/types/recovery.rs
```diff
@@ -0,0 +1,201 @@
...
+
+/// Recovery message data.
+///
+/// The message data can either be a string message that is first hashed
+/// according to EIP-191 and then recovered based on the signature or a
+/// precomputed hash.
+#[derive(Clone, Debug, PartialEq)]
+pub enum RecoveryMessage {
+    /// Message string
+    String(String),
```
tomusdrw: Shouldn't that be `Data(Bytes)`? I think you can sign arbitraty (non-utf8) bytes.

tomusdrw: I see that we have restricted hashing to `AsRef<str>` anyway, so it's good as is.

reuvenpo: Wait a second, that's a good point - shouldn't this crate allow hashing arbitrary data?

tomusdrw: Yeah, feel free to do a follow up PR on this, or maybe @nlordell is interested :)

nlordell: Looks like @reuvenpo beat me to it!

PR #287, opened 2019-11-20
Custom signing methods

reuvenpo:
This is a follow up PR to the ideas in [this comment](https://github.com/tomusdrw/rust-web3/pull/279#issuecomment-554799367) on #279. 

I define equivalents of methods that call functions on secret contracts, but i sign the transactions locally instead of letting the ethereum node sign them for me with an unlocked account.

The first commit addresses the issue pointed out in [this comment](https://github.com/tomusdrw/rust-web3/pull/279#discussion_r347868904) (Signing should accept arbitrary binary data rather than only UTF-8 data.)

### DIFF
--- a/src/api/accounts.rs
+++ b/src/api/accounts.rs
@@ -54,12 +54,14 @@ impl<T: Transport> Accounts<T> {
     /// using keccak256.
     pub fn hash_message<S>(&self, message: S) -> H256
     where
-        S: AsRef<str>,
+        S: AsRef<[u8]>,
     {
         let message = message.as_ref();
-        let eth_message = format!("\x19Ethereum Signed Message:\n{}{}", message.len(), message);
 
-        eth_message.as_bytes().keccak256().into()
+        let mut eth_message = Vec::from(format!("\x19Ethereum Signed Message:\n{}", message.len()).as_bytes());
+        eth_message.extend_from_slice(message);
+
+        eth_message.keccak256().into()
     }
 
     /// Sign arbitrary string data.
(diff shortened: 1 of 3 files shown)


tomusdrw:
Can you fix the formatting issue?

reuvenpo:
I fixed the formatting.
Please note that I opened this PR to implement the methods on Contract that my team needed, and i wrote these first few commits to build up to this, but this morning we de-prioritized this task and will sadly not complete it :/
Feel free to either merge this directly, or continue the work this PR was started for.

---
My only note would be that the fields of `web3::types::signed::{SignedData, SignedTransaction}` should be set to `pub(crate)` and methods on those types must be implemented to expose the values of these fields as readonly. It does not currently make sense that a user can have an instance of these types, and then write to it. That would break the invariant this type supposedly provides. I was gonna do this myself but i lost myself in a rabbit hole trying to understand the s, r, v fields. It should be documented which exact form of the v field is used in each one of the structs here, and perhaps even provide wrapper types for different v formats, along with From implementations between them. Again, it's hard to understand from the code as it is, what values can be expected in this field, and if it's even consistent between instances of the type.

tomusdrw:
Thanks for working on this! I feel that we don't really give any guarantees about the types, you can still get the error when you try to recover a signature from them, and being able to construct them outside is probably useful too, so I'll keep it as is.

reuvenpo:
what do you mean by "the error"?

tomusdrw:
`@reuvenpo This error for instance: https://github.com/tomusdrw/rust-web3/blob/355f0929ecf3a6ed1baa090f743d22f896c9b403/src/api/accounts.rs#L122

reuvenpo:
Ah, i see your point. Very well 👍

PR #291, opened 2019-12-02
estimate_gas: do not serialize null block number

### LINKED ISSUE #290: Estimate Gas Not Working With Geth
`Eth::estimate_gas` does not seem to work when connected to a geth node. From what I can gather this is because ethereum/go-ethereum#2586 where geth does not implement the documented Web3 specification for `eth_estimateGas`.

According to the [specification](https://github.com/ethereum/wiki/wiki/JSON-RPC#eth_estimategas), `eth_estimateGas` expects the same parameters as `eth_call`, that is (in `rust-web3` terms) a  `CallRequest` and a `BlockNumber`. However, geth seems to only accept a `CallRequest` and with fail when a `BlockNumber` is specified.

This is a problem when using any geth node (such as when using Infura). The following code illustrates this problem:

```rust
use futures::future::Future;
use serde_json::json;
use std::env;
use web3::api::Web3;
use web3::transports::Http;
use web3::types::CallRequest;
use web3::Transport;

fn main() {
    let (eloop, http) = {
        let project_id = env::var("INFURA_PROJECT_ID").unwrap();
        Http::new(&format!("https://rinkeby.infura.io/v3/{}", project_id)).unwrap()
    };
    eloop.into_remote();
    let web3 = Web3::new(http.clone());

    println!(
        "Client version: {:?}",
        web3.web3().client_version().wait().unwrap()
    ); // Client version: "Geth/v1.9.7-omnibus-cc8122b7-20191122/linux-amd64/go1.13.4"

    let gas = web3
        .eth()
        .estimate_gas(
            CallRequest {
                from: None,
                to: "000102030405060708090a0b0c0d0e0f10111213".parse().unwrap(),
                gas: None,
                gas_price: None,
                value: None,
                data: None,
            },
            None,
        )
        .wait();
    println!("{:?}", gas); // Err("too many arguments, want at most 1")

    let gas = http
        .execute(
            "eth_estimateGas",
            vec![json!({
                "to": "0x000102030405060708090a0b0c0d0e0f10111213",
            })],
        )
        .wait();
    println!("{:?}", gas); // Ok("0x5208") == 21000 == BASE_GAS_PRICE (as expected)
}
```

I suggest doing as `web3.py` does (ethereum/web3.py#1046) and only send the `CallRequest` parameter on a `eth_estimateGas` request when the specified `BlockNumber` is `None`. This works with other nodes such as parity and development nodes such as Ganache so it shouldn't impact `rust-web3` users while allowing them to successfully call `eth_estimateGas` when connected to geth nodes.

This feels icky as it requires `rust-web3` to work around an issue that is arguably a geth bug, but seeing as ethereum/go-ethereum#2586 has been open since 2016 I think it is a worthwhile workaround (especially since geth is widely used, and used with Infura).

vorot93:
Do not serialise second argument (block number) if it's null. This is a workaround for ethereum/go-ethereum#2586. Fixes #290.

### DIFF
--- a/src/api/eth.rs
+++ b/src/api/eth.rs
@@ -73,12 +73,9 @@ impl<T: Transport> Eth<T> {
     pub fn estimate_gas(&self, req: CallRequest, block: Option<BlockNumber>) -> CallFuture<U256, T::Out> {
         let req = helpers::serialize(&req);
 
-        let args = {
-            if let Some(block) = block {
-                vec![req, helpers::serialize(&block)]
-            } else {
-                vec![req]
-            }
+        let args = match block {
+            Some(block) => vec![req, helpers::serialize(&block)],
+            None => vec![req],
         };
 
         CallFuture::new(self.transport.execute("eth_estimateGas", args))
--- a/src/api/eth.rs
+++ b/src/api/eth.rs
@@ -463,6 +463,17 @@ mod tests {
   );
 
     rpc_test! (
+    Eth:estimate_gas:for_block, CallRequest {
+      from: None, to: Address::from_low_u64_be(0x123),
+      gas: None, gas_price: None,
+      value: Some(0x1.into()), data: None,
+    }, Some(0x123.into())
+    =>
+    "eth_estimateGas", vec![r#"{"to":"0x0000000000000000000000000000000000000123","value":"0x1"}"#, r#""0x123""#];
+    Value::String("0x123".into()) => 0x123
+  );
+
+    rpc_test! (
     Eth:gas_price => "eth_gasPrice";
     Value::String("0x123".into()) => 0x123
   );
--- a/src/api/eth.rs
+++ b/src/api/eth.rs
@@ -72,9 +72,16 @@ impl<T: Transport> Eth<T> {
     /// Call a contract without changing the state of the blockchain to estimate gas usage.
     pub fn estimate_gas(&self, req: CallRequest, block: Option<BlockNumber>) -> CallFuture<U256, T::Out> {
         let req = helpers::serialize(&req);
-        let block = helpers::serialize(&block.unwrap_or(BlockNumber::Latest));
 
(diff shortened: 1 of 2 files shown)


src/api/eth.rs
```diff
@@ -72,9 +72,16 @@ impl<T: Transport> Eth<T> {
     /// Call a contract without changing the state of the blockchain to estimate gas usage.
     pub fn estimate_gas(&self, req: CallRequest, block: Option<BlockNumber>) -> CallFuture<U256, T::Out> {
         let req = helpers::serialize(&req);
-        let block = helpers::serialize(&block.unwrap_or(BlockNumber::Latest));
 
-        CallFuture::new(self.transport.execute("eth_estimateGas", vec![req, block]))
+        let args = {
```
tomusdrw: Minor style nitpick:
```rust
let args = match block {
  Some(block) => vec![req, helpers::serialize(&block)],
  None => vec![req],
};
```
would be my preferred style.

src/contract/mod.rs
```diff
@@ -362,7 +362,7 @@ mod tests {
         };
 
         // then
-        transport.assert_request("eth_estimateGas", &["{\"data\":\"0x06fdde03\",\"from\":\"0x0000000000000000000000000000000000000005\",\"to\":\"0x0000000000000000000000000000000000000001\"}".into(), "\"latest\"".into()]);
+        transport.assert_request("eth_estimateGas", &["{\"data\":\"0x06fdde03\",\"from\":\"0x0000000000000000000000000000000000000005\",\"to\":\"0x0000000000000000000000000000000000000001\"}".into()]);
```
tomusdrw: Can you add a test with the block number set as well?

vorot93: Added in `Eth` transport tests.

src/api/eth.rs
vorot93: Fixed

PR #313, opened 2020-01-27
Adds impl Tokonizable for Bytes

chabashilah:
This is for using Bytes as a return value of `query` of `web3::contract::Contract`.


tomusdrw:
Thanks! Closes #314

PR #318, opened 2020-01-29
Implement `parity` namespace

gulitsky:
Do it mostly for very handy `parity_call` [method](https://github.com/paritytech/wiki/blob/master/JSONRPC-parity-module.md#parity_call) that allowing to run sequential `eth_call`s in one request. Also i'm tried to implement a batch version of `web3::contract::Contract::query` method using `parity_call`, but have not yet figured out how.

### DIFF
--- a/src/api/parity.rs
+++ b/src/api/parity.rs
@@ -27,7 +27,7 @@ impl<T: Transport> Namespace<T> for Parity<T> {
 impl<T: Transport> Parity<T> {
     /// Sequentially call multiple contract methods in one request without changing the state of the blockchain.
     pub fn call(&self, reqs: Vec<CallRequest>) -> CallFuture<Vec<Bytes>, T::Out> {
-        let req = helpers::serialize(&reqs);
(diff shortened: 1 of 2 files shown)


src/api/parity.rs
```diff
@@ -0,0 +1,77 @@
...
+        Parity { transport }
+    }
+
+    fn transport(&self) -> &T {
+        &self.transport
+    }
+}
+
+impl<T: Transport> Parity<T> {
+    pub fn call(&self, req: Vec<CallRequest>) -> CallFuture<Vec<Bytes>, T::Out> {
```
tomusdrw: Docs would be nice.

tomusdrw:
Can you please run `cargo fmt` on the code?
image

src/api/parity.rs
gulitsky: All looks good for now.

tomusdrw:
Perfect thanks!

Issue #335, opened 2020-03-31
Transaction index Uint different in Transaction and Log


brockelmore:
Transaction index for a transaction type is `Option<U128>` but for a log is `Option<U256>`. These should be consistent

tomusdrw:
PR welcome!

brockelmore:
Do you have a preference if its one way or another? I lean toward U256 but dont really care

tomusdrw:
it's `index` so it means it's bound by regular computer memory anyway, I would even go as far as making it `U64` if that type is available. There should be `U64 -> U128 -> U256` conversions in place anyway, right?

brockelmore:
On log, its a straight `Option<U256>` - no Index. I will update Index to be type U64 and push log to be Index if that is what you think is best?

tomusdrw:
Sounds good!

tomusdrw:
Closed via #336

PR #336, opened 2020-03-31
Update Index type to be U64, make tx_index consistent

brockelmore:
per: #335


tomusdrw:
@brockelmore build failed (no idea why it's not reported here). Could you run rustfmt?
https://travis-ci.org/github/tomusdrw/rust-web3/jobs/669282956

PR #345, opened 2020-05-14
Support eth_call by block hash

### LINKED ISSUE #285: Support EIP-1898 - geth state-related RPC methods now support specifying block hashes as well as block numbers
I noticed that Geth 1.9.6 added support for specifying block hashes for some methods (https://github.com/ethereum/go-ethereum/pull/19491).

https://eips.ethereum.org/EIPS/eip-1898

In my code, I've been passing the block number through, but passing the block hash through will give consistent answers if a re-org happens in the middle of some calls.

I think functions that have `block: Option<BlockNumber>` in their args would need to change, but I haven't thought about exactly how yet.

https://github.com/tomusdrw/rust-web3/blob/f043b222744580bf4be043da757ab0b300c3b2da/src/api/eth.rs#L42

leoyvens:
This resolves #285 by having the block parameter to `call` be a `BlockId` instead of a `BlockNumber` (breaking change), and serializing the `BlockId` in the manner expected by EIP 1898.

### DIFF
--- a/src/types/block.rs
+++ b/src/types/block.rs
@@ -1,5 +1,5 @@
 use crate::types::{Bytes, H160, H2048, H256, H64, U256, U64};
-use serde::{Deserialize, Serialize, Serializer};
+use serde::{ser::SerializeStruct, Deserialize, Serialize, Serializer};
 
 /// The block header type returned from RPC calls.
 #[derive(Debug, Clone, PartialEq, Deserialize, Serialize)]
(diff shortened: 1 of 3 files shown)


src/types/block.rs
```diff
@@ -158,7 +158,11 @@ impl Serialize for BlockId {
         S: Serializer,
     {
         match *self {
-            BlockId::Hash(ref x) => serializer.serialize_str(&format!("0x{}", x)),
+            BlockId::Hash(ref x) => {
```
tomusdrw, line 161: Initially I thought this is going to break some of the existing APIs using `BlockId`, but it's not the case. In fact it seems that the fact that `Serialize` is implemented for `BlockId` is not used anywhere in the `rust-web3` codebase internally, so it's okay to change it.

tomusdrw:
Thanks a lot!

src/types/block.rs
leoyvens: Exactly.

leoyvens:
Thanks for the quick review!

PR #346, opened 2020-05-14
Implement Serde Deserialize for Call and Transaction Requests

HCastano:
I've got a crate that's using these types, and it would benefit from being able to deserialize them.

### DIFF
--- a/src/types/transaction_request.rs
+++ b/src/types/transaction_request.rs
@@ -99,24 +99,20 @@ mod tests {
 
     #[test]
     fn should_deserialize_call_request() {
-        let call_request = CallRequest {
-            from: None,
-            to: Address::from_low_u64_be(5),
-            gas: Some(21_000.into()),
-            gas_price: None,
-            value: Some(5_000_000.into()),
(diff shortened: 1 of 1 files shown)


src/types/transaction_request.rs
```diff
@@ -128,4 +150,30 @@ mod tests {
 }"#
         );
     }
+
+    #[test]
+    fn should_deserialize_transaction_request() {
+        let tx_request = TransactionRequest {
```
tomusdrw, line 156: There is already a test for serialization, I think for simplicity of the test I'd prefer to just use a raw JSON as `serialized`.

src/types/transaction_request.rs
```diff
@@ -97,6 +97,28 @@ mod tests {
         );
     }
 
+    #[test]
+    fn should_deserialize_call_request() {
+        let call_request = CallRequest {
```
tomusdrw, line 102: same here - let's just hardcode a string (or two) and check we deserialize it correctly.

PR #348, opened 2020-05-18
Make CallRequest's `to` field optional

HCastano:
According to the Ethereum Wiki [page on JSON-RPC](https://github.com/ethereum/wiki/wiki/JSON-RPC#eth_estimategas), the `eth_call` and `eth_estimateGas` calls use the same parameters, with the exception that all the fields for `eth_estimateGas` are optional.

Currently, `CallRequest`'s `to` field was not optional. This PR changes that, making it suitable for use with `eth_estimateGas`.

I have a question about the RPC test I added with a `None` value for `to`. I didn't know what `eth_estimateGas` would return in that case, so I left the response from the test I copied. It passes, but I'm not sure it should considering that `Value::String("0x123".into()) => 0x123` was in reference to the `to` address. What should this test check instead?

### DIFF
--- a/src/api/eth.rs
+++ b/src/api/eth.rs
@@ -462,7 +462,7 @@ mod tests {
 
     rpc_test! (
       Eth:call, CallRequest {
-        from: None, to: Address::from_low_u64_be(0x123),
+        from: None, to: Some(Address::from_low_u64_be(0x123)),
         gas: None, gas_price: None,
         value: Some(0x1.into()), data: None,
       }, None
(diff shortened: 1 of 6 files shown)


src/api/eth.rs
```diff
@@ -502,9 +502,21 @@ mod tests {
       Value::String("0x123".into()) => 0x123
     );
 
+    // NOTE: I don't know why this is passing
```
tomusdrw, line 505:
```rust
 rpc_test! (
     /// What you want to call (with a test name)
      Eth:estimate_gas:optional_to_addr, CallRequest {
        from: None, to: None,
        gas: None, gas_price: None,
        value: Some(0x1.into()), data: None,
      }, None
      =>
     // what should be received (JSON)
      "eth_estimateGas", vec![r#"{"value":"0x1"}"#];
      /// What should the function return (`Value::String`) and what we expect to receive (JSON)
      Value::String("0x123".into()) => 0x123
    );
```

tomusdrw: The test is correct :)

Issue #17, opened 2017-01-10
Generate structs from contract interfaces


tomusdrw:
Example:
```rust
#[cfg(test)]
mod tests {
  use api;
  use helpers::tests::TestTransport;
  use types::U256;

  contract! {
    contract Token {
        event Transfer(address indexed from, address indexed to, uint256 value);
        event Approval(address indexed owner, address indexed spender, uint256 value);

        function balanceOf(address _owner) constant returns (uint256 balance);
        function transfer(address _to, uint256 _value) returns (bool success);
        function transferFrom(address _from, address _to, uint256 _value) returns (bool success);
        function approve(address _spender, uint256 _value) returns (bool success);
        function allowance(address _owner, address _spender) constant returns (uint256 remaining);
    }
  }

  #[test]
  fn should_call_token_contract {
    // given
    let transport = helpers::tests::TestTransport::default();
    let eth = api::Eth::new(transport);
    let token = Token::at(1.into(), eth);

    // when
    let balance = token.balanceOf(5.into());

    // then
    assert_eq!(balance, U256::from(10));

  }
}
```

rphmeier:
This would be seriously awesome :)
IMO this would be more easily accomplished if we mangled the solidity syntax a bit so all input/output types are just Rust-named. Then the types can just be used directly via `IntoTokens`/`FromTokens` and we don't have to maintain any kind of mapping from `uint256` -> `U256` or `address` -> `Address`, etc.

rphmeier:
It also might make sense to generate _modules_ from such a macro to encapsulate the events as tuple-structs.

tomusdrw:
Partially addressable by `ethabi_derive`. Should add examples how to use it (after `ethabi_derive` supports  async calls/transacts)

tomusdrw:
We should rather use `ethabi_derive` and make it more ergonomic to call using `web3`, see #120

That3Percent:
I'm working on something similar here - [solidity-bindgen](https://github.com/graphprotocol/solidity-bindgen)

It's not yet well documented, or in any sort of publishable state but you can do something like `solidity_bindgen::contract_abis!("../contracts/build/abis");` where the path given is a relative path to a directory containing abis. It will spit out structs for each contract with an async method for each function with named, strongly-typed parameters and return values.

If you want good IDE support, you can use `cargo-expand` to spit out a file since rust-analyzer doesn't autocomplete well for proc-macro yet.

tomusdrw:
[...]

PR #349, opened 2020-05-27
Contract calldata api

kyledewy:
Hey I wanted to use the library to construct calldata for contract functions, but I wasn't able to find anything.

I just did a quick change to see if I could expose the api for this, but I'm happy to clean it up if it makes sense.

### DIFF
--- a/src/contract/mod.rs
+++ b/src/contract/mod.rs
@@ -100,20 +100,14 @@ impl<T: Transport> Contract<T> {
         Ok(Self::new(eth, address, abi))
     }
 
+    /// Get the underlying contract ABI.
+    pub fn abi(&self) -> &ethabi::Contract {
+        &self.abi
+    }
+
     /// Returns contract address
     pub fn address(&self) -> Address {
         self.address
-    }
-
-    /// Returns calldata for contract function call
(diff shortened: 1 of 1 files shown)


src/contract/mod.rs
```diff
@@ -105,37 +105,35 @@ impl<T: Transport> Contract<T> {
...
             .and_then(|function| function.encode_input(&params.into_tokens()))
-            .map(move |data| {
-                let Options {
-                    gas,
-                    gas_price,
-                    value,
-                    nonce,
-                    condition,
-                } = options;
+            .unwrap()
```
tomusdrw, line 116: The error is there for a reason, simply `unwrap`ing will cause panics in case the function is not found.

tomusdrw:
Awesome, cheers!

src/contract/mod.rs
kyledewy: Thanks for pointing me towards [#17](https://github.com/tomusdrw/rust-web3/issues/17) .. ethabi_derive looks really nice :ok_hand: 
Sorry about the unwrap() .. should have been an issue not a pull request :roll_eyes:

tomusdrw: No worries @kyledewy, thanks a lot for the contribution!

PR #360, opened 2020-06-26
Add support for wss

en:
Make rust-web3 work with an infura-like endpoint.
For example, `wss://mainnet.infura.io/ws/v3/{PROJECT_ID}`

### DIFF
--- a/src/transports/ws.rs
+++ b/src/transports/ws.rs
@@ -1,6 +1,7 @@
 //! WebSocket Transport
 
 use std::collections::BTreeMap;
+use std::marker::Unpin;
 use std::sync::{atomic, Arc};
 use std::{fmt, pin::Pin};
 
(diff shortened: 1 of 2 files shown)


src/transports/ws.rs
```diff
@@ -36,20 +40,85 @@ type BatchResult = error::Result<Vec<SingleResult>>;
 type Pending = oneshot::Sender<BatchResult>;
 type Subscription = mpsc::UnboundedSender<rpc::Value>;
 
+/// Stream, either plain TCP or TLS.
+enum MaybeTlsStream<S> {
+    /// Unencrypted socket stream.
+    Plain(S),
```
tomusdrw, line 46: We could use `pin_project` to avoid the `Unpin` bound, but it's already a separate issue to work on. See #361

PR #365, opened 2020-07-06
Isolate crypto stuff and avoid dealing with leaking `SecretKey`

tomusdrw:
Follow up on https://github.com/tomusdrw/rust-web3/pull/358#pullrequestreview-440205023

This PR re-organizes the code using crypto stuff into `signing.rs` module and hides `secp256k1` details behind the `Sign` trait.

The purpose of this is to avoid handling `SecretKey` and needing to zeroize or protect it at any point in time, rather this responsibility is passed to the users of the library.
Secondly this should make it easier to switch to a different (for instance `no-std` compatible) version of `secp256k1` (like `libsecp256k1`) in the future since `secp256k1` is not part of the public API any more.

This might be a little bit more cumbersome to use, we can possibly provide examples if that becomes an issue.

CC @That3Percent @cheme :)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -27,7 +27,6 @@ secp256k1 = { version = "0.17", features = ["recovery"] }
 serde = { version = "1.0.90", features = ["derive"] }
 serde_json = "1.0.39"
 tiny-keccak = { version = "2.0.1", features = ["keccak"] }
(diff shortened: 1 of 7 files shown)


tomusdrw:
Thanks for reviewing this PR, merging then.

PR #370, opened 2020-08-24
Resolving HTTP_PROXY variables if set.

### LINKED ISSUE #369: Connect through a proxy if environment variables are set.
It would be nice, if a connection was established through proxies as specified in the env vars: http_proxy, https_proxy, HTTP_PROXY or HTTPS_PROXY
It looks like hyper has support for it, but requires the client to read the env vars.

whizsid:
Fixes #369 .

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -52,7 +52,7 @@ tokio = { version = "0.2", features = ["full"] }
 [features]
 default = ["http-tls", "ws-tls-tokio"]
 http = ["hyper", "hyper-proxy", "url", "base64", "typed-headers"]
-http-tls = ["hyper-tls", "hyper-proxy", "native-tls", "http", "typed-headers"]
+http-tls = ["hyper-tls", "hyper-proxy", url ,"native-tls", "http", "typed-headers"]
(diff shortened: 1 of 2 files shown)


Cargo.toml
```diff
@@ -49,8 +51,8 @@ tokio = { version = "0.2", features = ["full"] }
 
 [features]
 default = ["http-tls", "ws-tls-tokio"]
-http = ["hyper", "url", "base64"]
-http-tls = ["hyper-tls", "native-tls", "http"]
+http = ["hyper", "hyper-proxy", "url", "base64", "typed-headers"]
+http-tls = ["hyper-tls", "hyper-proxy", "url" ,"native-tls", "http", "typed-headers"]
```
tomusdrw, line 55: Since, `http` is a required feature of `http-tls` already, these other extra deps don't need to be listed explicitly.

```suggestion
http-tls = ["hyper-tls", "native-tls", "http"]
```

tomusdrw (changes requested):
Looks really good! Just a tiny grumble regarding unnecessary features.

tomusdrw (approved):

tomusdrw:
Thanks a lot!

PR #372, opened 2020-08-30
remove unused generic from call_with_confirmations

sinkingsugar:
`<P>` is not used anywhere as we use `impl Tokenize` instead.
This prevents compilation.


tomusdrw (approved):
Thanks!

PR #374, opened 2020-09-10
Fix net_version doc string

e00E:
See https://eth.wiki/json-rpc/API#net_version . This method returns the network id, not the protocol version (which is returned by `eth_protocolVersion`).


tomusdrw (approved):
thanks!

PR #377, opened 2020-09-24
Update implementation of Tokenizable for Vec<u8>

ducthotran2010:
The contract call could return `uint8[]` instead of `bytes`, but `Vec<u8>` implementation just supports `bytes`.

### DIFF
--- a/src/contract/tokens.rs
+++ b/src/contract/tokens.rs
@@ -1,7 +1,7 @@
 //! Contract Functions Output types.
 
 use crate::contract::error::Error;
-use crate::types::{Address, Bytes, H256, U128, U256};
+use crate::types::{Address, Bytes, BytesArray, H256, U128, U256};
 use arrayvec::ArrayVec;
 use ethabi::Token;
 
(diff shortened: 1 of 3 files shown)


src/contract/tokens.rs
```diff
@@ -298,9 +298,11 @@ tokenizable_item! {
 impl Tokenizable for Vec<u8> {
     fn from_token(token: Token) -> Result<Self, Error> {
         match token {
-            Token::Bytes(data) => Ok(data),
-            Token::FixedBytes(data) => Ok(data),
-            other => Err(Error::InvalidOutputType(format!("Expected `bytes`, got {:?}", other))),
+            Token::Bytes(data) | Token::FixedBytes(data) => Ok(data),
+            Token::Array(tokens) | Token::FixedArray(tokens) => {
```
tomusdrw, line 302: The problem with this is that you will decode `Vec<u8>` from `Token::Array`, but encoding (`into_tokens`) will always return `Token::Bytes` which is quite misleading.

The split is done on purpose currently (also a bit due to Rust limitations). I think the best way forward would be to introduce a type that looks like this:
```rust
pub struct BytesArray(pub Vec<u8>);
impl Tokenizable for BytesArray {
    fn from_token(token: Token) -> Result<Self, Error> {
        match token {
            Token::FixedArray(tokens) | Token::Array(tokens) => {
                Ok(Self(tokens.into_iter().map(Tokenizable::from_token).collect()?))
            }
            other => Err(Error::InvalidOutputType(format!("Expected `Array`, got {:?}", other))),
        }
    }

    fn into_token(self) -> Token {
        Token::Array(self.into_iter().map(Tokenizable::into_token).collect())
    }
}
```

Unless `Token::Bytes` and `Token::Array` have the same ethabi encoding and there is no confusion here (I really don't remember).

Also please add tests.

tomusdrw (changes requested):

src/types/bytes_array.rs
```diff
@@ -0,0 +1,5 @@
+use serde::{Deserialize, Serialize};
+
+/// Array of bytes wrapper
```
tomusdrw, line 3:
```suggestion
/// A wrapper type for array of bytes.
/// 
/// Implements `Tokenizable` so can be used to retrieve data from `Solidity` contracts returning `byte8[]`.
```

tomusdrw (commented):
Awesome, thanks! tiny doc improvement

tomusdrw (approved):

tomusdrw:
Thank you!

src/types/bytes_array.rs
ducthotran2010: Nice

PR #378, opened 2020-09-24
Support canonical version of the signature

ducthotran2010:
https://github.com/ethereum/go-ethereum/issues/2053
https://github.com/ethereumjs/ethereumjs-util/blob/584cc821b5f89980d1c1f87a6a4fe2610aa5c5a1/src/signature.ts#L67-L90

### DIFF
--- a/src/types/recovery.rs
+++ b/src/types/recovery.rs
@@ -49,7 +49,12 @@ impl Recovery {
             return Err(ParseSignatureError);
         }
 
-        let v = bytes[64];
+        let mut v = bytes[64];
+
+        if v < 27 {
+            v += 27;
+        }
+
         let r = H256::from_slice(&bytes[0..32]);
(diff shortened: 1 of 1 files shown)


tomusdrw:
Is this strictly required for compatibility with some modern clients? We enforce the `Electrum` (27,28) notation everywhere in the code and I'd rather prefer to be strict about it than to simply "correct" invalid `v` values.

Instead of this I'd be more willing to add a companion method for `from_raw_signature` (say `from_raw_signature_canon`) and only use it in case we know the client is sending INVALID data.

tomusdrw:
@ducthotran2010 I'm closing this due to inactivity. As mentioned earlier I'm down to include an alternative function to import non-common/non-electrum notation of signatures, so please re-open the PR if you wish to add it this way.

Issue #381, opened 2020-10-02
txpool Namespace support


ksato9700:
Is there a plan to support txpool namespace methods? If not, can I work on it?
https://geth.ethereum.org/docs/rpc/ns-txpool

PR #382, opened 2020-10-03
add txpool namespace support

ksato9700:
This is a PR for the issue #381. Please review and merge it if you like.

### DIFF
--- a/src/types/txpool.rs
+++ b/src/types/txpool.rs
@@ -1,27 +1,50 @@
-use std::collections::HashMap;
-use crate::types::{Transaction};
+use std::collections::BTreeMap;
+use crate::types::{Transaction, Address};
 use serde::{Deserialize, Serialize};
 use crate::types::U64;
 
 /// Transaction Pool Content Info
+///
+/// The content inspection property can be queried to list the exact details of all
+/// the transactions currently pending for inclusion in the next block(s), as well
+/// as the ones that are being scheduled for future execution only.
+///
+/// See [here](https://geth.ethereum.org/docs/rpc/ns-txpool#txpool_content) for more details
+///
 #[derive(Debug, Default, Clone, PartialEq, Deserialize, Serialize)]
 pub struct TxpoolContentInfo {
     /// pending tx
-    pub pending: HashMap<String, HashMap<String, Transaction>>,
+    pub pending: BTreeMap<Address, BTreeMap<String, Transaction>>,
     /// queued tx
-    pub queued: HashMap<String, HashMap<String, Transaction>>,
+    pub queued: BTreeMap<Address, BTreeMap<String, Transaction>>,
 }
 
 /// Transaction Pool Inspect Info
+///
+/// The inspect inspection property can be queried to list a textual summary
+/// of all the transactions currently pending for inclusion in the next block(s),
+/// as well as the ones that are being scheduled for future execution only.
+/// This is a method specifically tailored to developers to quickly see the
+/// transactions in the pool and find any potential issues.
+///
+/// See [here](https://geth.ethereum.org/docs/rpc/ns-txpool#txpool_inspect) for more details
+///
 #[derive(Debug, Default, Clone, PartialEq, Deserialize, Serialize)]
 pub struct TxpoolInspectInfo {
     /// pending tx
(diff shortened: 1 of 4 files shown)


src/types/txpool.rs
```diff
@@ -0,0 +1,204 @@
+use std::collections::HashMap;
+use crate::types::{Transaction};
+use serde::{Deserialize, Serialize};
+use crate::types::U64;
+
+/// Transaction Pool Content Info
+#[derive(Debug, Default, Clone, PartialEq, Deserialize, Serialize)]
+pub struct TxpoolContentInfo {
+    /// pending tx
+    pub pending: HashMap<String, HashMap<String, Transaction>>,
```
tomusdrw, line 10: Perhaps better to use `BTreeMap` here to have the keys deterministically ordered? With `HashMap` the results would be a bit random. Also we should be using more strict types than `String`. I believe it's transaction hash and the nonce, right?
```suggestion
    pub pending: BTreeMap<H256, BTreeMap<U256, Transaction>>,
```

src/types/txpool.rs
```diff
@@ -0,0 +1,204 @@
...
+    /// queued tx
+    pub queued: HashMap<String, HashMap<String, Transaction>>,
```
tomusdrw, line 12:
```suggestion
    pub queued: BTreeMap<H256, BTreeMap<U256, Transaction>>,
```

src/types/txpool.rs
```diff
@@ -0,0 +1,204 @@
...
+}
+
+/// Transaction Pool Inspect Info
+#[derive(Debug, Default, Clone, PartialEq, Deserialize, Serialize)]
+pub struct TxpoolInspectInfo {
+    /// pending tx
+    pub pending: HashMap<String, HashMap<String, String>>,
```
tomusdrw, line 19:
```suggestion
    pub pending: BTreeMap<H256, BTreeMap<U256, String>>,
```

src/types/txpool.rs
```diff
@@ -0,0 +1,204 @@
...
+    /// queued tx
+    pub queued: HashMap<String, HashMap<String, String>>,
```
tomusdrw, line 21:
```suggestion
    pub queued: BTreeMap<H256, BTreeMap<U256, String>>,
```

src/types/txpool.rs
```diff
@@ -0,0 +1,204 @@
...
+/// Transaction Pool Content Info
+#[derive(Debug, Default, Clone, PartialEq, Deserialize, Serialize)]
+pub struct TxpoolContentInfo {
+    /// pending tx
+    pub pending: HashMap<String, HashMap<String, Transaction>>,
+    /// queued tx
+    pub queued: HashMap<String, HashMap<String, Transaction>>,
+}
+
+/// Transaction Pool Inspect Info
```
tomusdrw, line 15: Can we expand on the docs or at least point to documentation of these structs? I'd be more interested in the `String` format of the transaction inspection info - i.e. what things can be found there and can we actually parse it?

src/types/txpool.rs
```diff
@@ -0,0 +1,204 @@
...
+        let _: TxpoolInspectInfo = serde_json::from_str(txpool_inspect_str).unwrap();
+    }
+
+    #[test]
+    fn should_deserialize_txpool_status() {
+        let txpool_status_str = r#"{
+            "pending":"0x23",
+            "queued":"0x20"
+        }"#;
+        let _: TxpoolStatus = serde_json::from_str(txpool_status_str).unwrap();
```
tomusdrw, line 202: Re-serialization test would be cool, after switching to `BTreeMap` you will get consistent results.

tomusdrw (commented):
Looks really good for a start, thanks! I'd like to see a bit more strict typing and some documentation pointers.

ksato9700:
@tomusdrw Thank you for the review comments. I made changes as follows.
* use `BTreeMap` instead of `HashMap`
* use `Address` for the type of outer key
* add some comments with link to a spec document
* add re-serialization tests

Note that I keep using `String` for the inner key type. As you said, it is nonce, but strangely enough, decimal number format is used here instead of hex.

tomusdrw (approved):
looks awesome, thanks!

Issue #361, opened 2020-06-26
Get rid of Unpin requirements.


tomusdrw:
We should attempt to rewrite the library to avoid having transports require `Unpin` outputs.

PR #387, opened 2020-10-16
Convert transaction signing to `async`

e00E:
Converting the manual Future implementation to async makes the code
simpler and no longer require Unpin.

part of  #361


tomusdrw (approved):
That's amazing :heart_eyes:, thank you so much!

e00E:
Are you sure you want to bump version immediately? To be clear, this code does not yet have outside observable effect but it is a part of several changes that will be required to remove Unpin.

tomusdrw:
> Are you sure you want to bump version immediately?

Bumping the version, just so that I don't forget to do it later on. Different return types is a breaking change already, so I'd prefer to indicate that the git version is different than what's on crates.io. Totally fine with merging more changes before the actual release.

e00E:
Makes sense :+1:

PR #388, opened 2020-10-16
Convert contract::deploy to async

e00E:
part of #361

Note that I have changed `execute` to return a single Result instead of nested Results by extending `deploy::Error` with AbiError.


tomusdrw (approved):
Thanks!

PR #394, opened 2020-10-23
Convert confirm and eth_filter to async

### LINKED ISSUE #393: FilterStream violates Future contract by potentially polling a completed future
I noticed this in one of the async rewrites (which also fixes the problem). This issue is for documentation purposes.

The line
https://github.com/tomusdrw/rust-web3/blob/2c33873488cbf9a27bf38d665138b227521bf51a/src/api/eth_filter.rs#L83
applies the `?` operator to the completed result of the inner future. If this result is ok then there is no bug. If the result is Error we forward the error as part of our stream which means that we don't change `self.state` and that the future stays. If the stream is polled again it will poll the inner future again even though it has already completed (with the previous error) which is not allowed.

e00E:
It would be good if someone could double check that I did not accidentally introduce any logic changes when converting the state machine manual future implementations to async.

I've marked some problems that I noticed in the existing code while converting it. This PR doesn't fix them but I put a comment there to mark them.

Fixes #393

### DIFF
--- a/src/confirm.rs
+++ b/src/confirm.rs
@@ -1,17 +1,10 @@
 //! Easy to use utilities for confirmations.
 
-use futures::stream::Skip;
-use futures::{
-    task::{Context, Poll},
-    Future, FutureExt, StreamExt,
-};
-use std::pin::Pin;
-use std::time::Duration;
-
-use crate::api::{CreateFilter, Eth, EthFilter, FilterStream, Namespace};
-use crate::helpers::CallFuture;
+use crate::api::{Eth, EthFilter, Namespace};
 use crate::types::{Bytes, TransactionReceipt, TransactionRequest, H256, U64};
 use crate::{error, Transport};
+use futures::{Future, StreamExt};
+use std::time::Duration;
 
 /// Checks whether an event has been confirmed.
 pub trait ConfirmationCheck {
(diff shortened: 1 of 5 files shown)


src/confirm.rs
```diff
@@ -34,279 +27,93 @@ where
...
+    // TODO: The stream should have additional checks:
+    // * We should not continue calling next on a stream that has completed (has returned None). We expect this to never
+    //   happen for the blocks filter but to be safe we should handle this case for example by `fuse`ing the stream or
+    //   erroring when it does complete.
+    // * We do not handle the case where the stream returns an error which means we are wrongly counting it as a
+    //   confirmation.
+    let filter_stream = filter.stream(poll_interval).skip(confirmations);
+    futures::pin_mut!(filter_stream);
+    loop {
+        let _ = filter_stream.next().await;
```
e00E, line 53: This bug already existed so this PR doesn't make it worse.

src/confirm.rs
```diff
@@ -34,279 +27,93 @@ where
...
+        let eth = eth.clone();
+        wait_for_confirmations(eth, eth_filter, poll_interval, confirmations, confirmation_check).await?;
     }
+    // TODO: We should remove this `expect`. No matter what happens inside the node, this shouldn't be a panic.
+    // Should convert this into an some error. InternalError probably makes sense but then we cannot provide the error
+    // message. Maybe make new error for this? Or allow InternalError to carry a message.
+    let receipt = eth
+        .transaction_receipt(hash)
+        .await?
+        .expect("receipt can't be null after wait for confirmations; qed");
```
e00E, line 87: This bug already existed so this PR doesn't make it worse.

src/confirm.rs
tomusdrw: I presume this behaviour was already broken before the refactor, right? Can we please log this as a separate issue and add an issue number to the `TODO` above?

src/confirm.rs
tomusdrw: Maybe let's put the responsibility to `unwrap` on the user? I.e. returning `Option<TransactionReceipt>` here?

tomusdrw (commented):
Looks really nice! I didn't find anything suspicious, only couple of minor things to fix, thanks!

src/confirm.rs
e00E: Yes, this PR doesn't change the behavior (bug was already there). I don't want to fix it in this PR. Added issue link.

src/confirm.rs
e00E: Like above the issue was already there and I don't want to fix it in this PR. Added issue link.

tomusdrw (approved):
Great, thanks again!

PR #395, opened 2020-10-23
Fix clippy warnings

e00E:
Fixes all warnings emitted by clippy because all of them were
reasonable.

When I was touching imports I removed some empty lines because rustfmt
already orders imports.


tomusdrw (approved):
awesome, thanks!

PR #399, opened 2020-10-30
Remove Unpin requirement from Transport

e00E:
This is based on the other PR for eth_subscribe so it includes that commit too. After making use of pin-project there I realized I could do this directly for CallFuture and thus make Transport no longer require Unpin even without removing all of the uses of CallFuture. There are still uses of Unpin in a couple of tests, http.rs and ws.rs . It looks like those are benign because everything still compiles but I am a little bit wary that this might have some unexpected consequence.

### DIFF
--- a/src/transports/either.rs
+++ b/src/transports/either.rs
@@ -1,6 +1,10 @@
 //! A strongly-typed transport alternative.
 
 use crate::{api, error, rpc, BatchTransport, DuplexTransport, RequestId, Transport};
+use futures::{
+    future::{BoxFuture, FutureExt},
+    stream::{BoxStream, StreamExt},
+};
 
 /// A wrapper over two possible transports.
 ///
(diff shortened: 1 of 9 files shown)


src/transports/either.rs
```diff
@@ -20,10 +24,10 @@ impl<A, B, AOut, BOut> Transport for Either<A, B>
 where
     A: Transport<Out = AOut>,
     B: Transport<Out = BOut>,
-    AOut: futures::Future<Output = error::Result<rpc::Value>> + Unpin + 'static,
-    BOut: futures::Future<Output = error::Result<rpc::Value>> + Unpin + 'static,
+    AOut: futures::Future<Output = error::Result<rpc::Value>> + 'static + Send,
```
e00E, line 27: I *think* previously EitherTransport did not require Send but now it does because of BoxFuture. I wasn't able to get the types to work out otherwise. Technically one could duplicate this code and use LocalBoxFuture to get rid of the Send requirement.

tomusdrw: IMHO acceptable trade-off for async code for now.

tomusdrw (approved):

Issue #134, opened 2018-06-29
Add a public export for web3::helpers::TestTranport


adklempner:
I'd like to use the mocked `TestTransport` struct for testing a crate that depends on `rust-web3`.

Is there already a way of doing this? If not, what would be the best way to export it? Would be happy to open a PR for this.

tomusdrw:
Sounds good to me, should probably be feature-gated like other transports.

creese:
I tried `use web3::helpers::tests::TestTransport` but I'm getting an unresolved import error. Why doesn't this work?

tomusdrw:
The entire `tests` module is gated with `#[cfg(test)]` which means it's only compiled when tests for `web3` crate are running (the tests running on your crate that depends on `web3` does not inlcude this code).

My earlier suggestion was to extract `TestTransport` to `web3::transports` module and expose it from there, adding a Cargo `feature` for the test transport (just like we have `http` and `ws` transports now.

marmistrz:
What do you think about switching `Rc<RefCell<_>>` to `Arc<RwLock<_>>` so that `TestTransport` implements `Sync + Send`?

PR #402, opened 2020-11-05
Extract TestTransport to web3::transports

creese:


src/transports/mod.rs
```diff
@@ -15,6 +15,9 @@ pub mod ws;
 #[cfg(feature = "ws")]
 pub use self::ws::WebSocket;
 
+#[cfg(feature = "test")]
```
tomusdrw, line 18:
```suggestion
#[cfg(any(feature = "test", test))]
```

Cargo.toml
```diff
@@ -48,10 +48,11 @@ tokio = { version = "0.2", features = ["full"] }
 async-std = { version = "1.5.0", features = ["attributes"] }
 
 [features]
-default = ["http", "ws", "http-tls", "ws-tls"]
-http = ["hyper", "url", "base64"]
+default = ["http", "ws", "http-tls", "ws-tls", "test"]
+http = ["hyper", "url", "base64", "test"]
```
tomusdrw, line 52:
```suggestion
default = ["http", "ws", "http-tls", "ws-tls"]
http = ["hyper", "url", "base64"]
```

Let's disable `test` by default, but change `cfg` stuff to enable it for tests as well to make sure they work without enabling any extra features.

tomusdrw (commented):
Looks great, thanks! Two small changes I'd like to see though.

Cargo.toml
creese: I'm wondering why we wouldn't want this by default since we nearly always want unit tests. Is there another way to write unit tests that I haven't found?

creese: Or that what you meant, by default, the feature is enabled but only for tests?

src/transports/mod.rs
creese: Can you explain what this does?

tomusdrw: It enables the module either when we run tests for `web3` or when `test` feature for the crate is enabled.

Cargo.toml
tomusdrw: Yes, that's why I suggested the `#[cfg(any` stuff. Even though it might be useful for external developers to have the test transport, they can enable it only for `dev-dependencies`. AFAIK current cargo does not conflate the features between these deps any more, so it would be cleaner.

tomusdrw:
Closes #134

tomusdrw (approved):

tomusdrw:
Thanks!

PR #403, opened 2020-11-06
Remove more instances of Unpin

e00E:
Not required for `Transport` but I thought I'd clean up all of the remaining Unpin instances (except for `ws.rs`).

### DIFF
--- a/src/transports/http.rs
+++ b/src/transports/http.rs
@@ -229,10 +229,13 @@ impl<T> Response<T> {
     }
 }
 
+// We can do this because `hyper::client::ResponseFuture: Unpin`.
+impl<T> Unpin for Response<T> {}
+
 impl<T, Out> Future for Response<T>
 where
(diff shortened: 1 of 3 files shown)


src/transports/http.rs
```diff
@@ -229,10 +229,13 @@ impl<T> Response<T> {
     }
 }
 
+// We can do this because `hyper::client::ResponseFuture: Unpin`.
+impl<T> Unpin for Response<T> {}
```
tomusdrw, line 233: What about `T`?

e00E: If we trust the compiler then this must be okay because it does compile. I think this is because we never construct a `Pin<T>` so whether T is Unpin doesn't matter. Although I am not sure why the compiler doesn't automatically implement Unpin in this case.

tomusdrw (approved):

PR #405, opened 2020-11-13
Update block.rs for pending block

yq0810:


src/types/block.rs
```diff
@@ -14,7 +14,7 @@ pub struct BlockHeader {
     pub uncles_hash: H256,
     /// Miner/author's address.
     #[serde(rename = "miner")]
-    pub author: H160,
+    pub author: Option<H160>,
```
tomusdrw, line 17: It's really weird that `author/miner` is not set, I can't find any evidence of this change. Since this is an edge case and I can imagine anyone really wanting to handle the `None` case, let's please use default value instead of `Option`.

```suggestion
    #[serde(default)]
    pub author: H160,
```

src/types/block.rs
```diff
@@ -63,7 +63,7 @@ pub struct Block<TX> {
...
```
tomusdrw, line 66:
```suggestion
    #[serde(default)]
    pub author: H160,
```

tomusdrw (changes requested):

PR #406, opened 2020-11-15
Fix Eth::uncle

vorot93:


tomusdrw:
The API docs say the return type should be `Block`, but without individual transactions (so just hashes), see here:
https://eth.wiki/json-rpc/API#returns-33

Open Ethereum returns a `RichBlock` struct as well, not just a `Header`:
https://github.com/openethereum/openethereum/blob/main/rpc/src/v1/impls/eth.rs#L359

Can you provide some context for this change?

vorot93:
Exactly what you said, basically. I'm using this library against turbo-geth's `rpcdaemon` and IIRC ran into abscence of TX hashes.

tomusdrw (approved):
Thanks, I've changed the API a bit to avoid breaking backward compatibility. So there is `Eth::uncle_header` for TurboGeth compatibility.

PR #407, opened 2020-11-15
Merge RawTransactionDetails into Transaction

vorot93:


src/api/eth.rs
```diff
@@ -136,6 +136,24 @@ impl<T: Transport> Eth<T> {
         CallFuture::new(result)
     }
 
+    /// Get block details with full transaction objects.
+    pub fn block_with_raw_txs(&self, block: BlockId) -> CallFuture<Option<Block<RawTransactionDetails>>, T::Out> {
```
tomusdrw, line 140: I guess the are only few differences between `RawTransactionDetails` and `Transaction`:
- `from` being `Optional`
- `r,s,v` for signature details.

Given this, I think I'd rather be for unifying the two structs and leaving the fields `Optional`, what do you think?

tomusdrw: @vorot93 do you mind rebasing and merging `RawTransactionDetails` and `Transaction` by having all the fields as `Option`?

vorot93: done

tomusdrw (approved):

PR #411, opened 2020-11-20
Add support for EIP-234

GCdePaula:
# [EIP-234](https://eips.ethereum.org/EIPS/eip-234): Add `blockHash` to JSON-RPC filter options.

Small changes to `Filter` and `FilterBuilder`, to add support for EIP-234. Here's the related API docs:
https://eth.wiki/json-rpc/API#eth_getlogs

Cheers!

### DIFF
--- a/src/types/log.rs
+++ b/src/types/log.rs
@@ -82,6 +82,9 @@ pub struct Filter {
     /// To Block
     #[serde(rename = "toBlock", skip_serializing_if = "Option::is_none")]
     to_block: Option<BlockNumber>,
+    /// Block Hash
+    #[serde(rename = "blockHash", skip_serializing_if = "Option::is_none")]
+    block_hash: Option<H256>,
     /// Address
     #[serde(skip_serializing_if = "Option::is_none")]
     address: Option<ValueOrArray<H160>>,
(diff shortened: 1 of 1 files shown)


src/types/log.rs
```diff
@@ -102,16 +105,26 @@ pub struct FilterBuilder {
...
     }
 
     /// Sets to block
     pub fn to_block(mut self, block: BlockNumber) -> Self {
+        self.filter.block_hash = None;
         self.filter.to_block = Some(block);
         self
     }
 
+    /// Sets block hash
```
tomusdrw, line 120: Could you update documentation and mention that it's mutually exclusive with `from_block` and `to_block`?

Same with `from_block` and `to_block` docs.

tomusdrw (commented):
Looks good! I'd like to have the docs updated to reflect the implementation.

src/types/log.rs
GCdePaula: Sure!

tomusdrw (approved):

src/types/log.rs
tomusdrw: Thanks!

PR #414, opened 2020-11-28
Add EIP-1193 support

enolan:
I added support for EIP-1193 and got the library working in the browser. This will be easier to review commit-by-commit.

For an example of the new functionality in use, you can check out the `eip-1193-with-example` branch from my fork. To run: `wasm-pack build -- --no-default-features --features eip-1193 && cd www && npm install && npm run start`. Then open http://127.0.0.1:8080 in a browser with MetaMask installed. Reject the transaction when prompted unless you enjoy burning 1 ETH for no reason.

This is my first substantial Rust code, feedback is very appreciated.

### DIFF
--- a/src/api/accounts.rs
+++ b/src/api/accounts.rs
@@ -1,23 +1,15 @@
 //! Partial implementation of the `Accounts` namespace.
 
 use crate::api::Namespace;
-#[cfg(feature = "signing")]
-use crate::api::Web3;
-#[cfg(feature = "signing")]
-use crate::error;
 use crate::signing;
 #[cfg(feature = "signing")]
 use crate::signing::Signature;
 use crate::types::H256;
 #[cfg(feature = "signing")]
-use crate::types::{
-    Address, Bytes, Recovery, RecoveryMessage, SignedData, SignedTransaction, TransactionParameters, U256,
-};
+use crate::types::{Address, SignedTransaction, U256};
 use crate::Transport;
 #[cfg(feature = "signing")]
 use rlp::RlpStream;
-#[cfg(feature = "signing")]
-use std::convert::TryInto;
 
 /// `Accounts` namespace
 #[derive(Debug, Clone)]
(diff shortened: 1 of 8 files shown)


src/transports/eip_1193.rs
```diff
@@ -0,0 +1,189 @@
...
+struct RequestArguments {
+    method: String,
+    params: js_sys::Array,
+}
+
+#[wasm_bindgen]
+impl RequestArguments {
+    #[wasm_bindgen(getter)]
+    pub fn method(&self) -> String {
+        self.method.clone()
```
tomusdrw, line 182: Do we need to clone? Why can't we return just references?

src/transports/mod.rs
```diff
@@ -31,3 +31,6 @@ impl From<native_tls::Error> for crate::Error {
         crate::Error::Transport(format!("{:?}", err))
     }
 }
+
+#[cfg(feature = "eip-1193")]
```
tomusdrw, line 35: Oh, the feature gate is already here, so an extra one in the module is definitely not needed.

src/transports/eip_1193.rs
```diff
@@ -0,0 +1,189 @@
...
+use futures::future::LocalBoxFuture;
+use jsonrpc_core::types::request::{Call, MethodCall};
+use std::cell::RefCell;
+use std::collections::BTreeMap;
+use std::rc::Rc;
+use wasm_bindgen::prelude::*;
+
+/// EIP-1193 transport
+#[derive(Clone, Debug)]
+pub struct Eip1193 {
```
tomusdrw, line 22: To be fair I prefer this notation over one with dash or an underscore. So I'd prefer the feature and the enclosing module to be both called `eip1193` and that would align with the struct as well. I'm not super strong on this, so if the `-` notation is more correct, so be it.

tomusdrw (changes requested):
Thanks a lot for this! There is a couple of fixes I'd like to see, but functionality-wise it looks complete and extremely useful.

src/api/accounts.rs
```diff
@@ -31,6 +40,7 @@ impl<T: Transport> Namespace<T> for Accounts<T> {
 
 impl<T: Transport> Accounts<T> {
     /// Gets the parent `web3` namespace
+    #[cfg(feature = "signing")]
```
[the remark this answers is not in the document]
enolan: I did as you said with a conditional `impl` block. `web3` *is* totally OK, but it's unused with the signing stuff turned off, so the compiler complains.

src/transports/eip_1193.rs
enolan: You can't return references to JS-land unfortunately. There's no way for the borrow checker to see into JavaScript.

src/transports/eip_1193.rs
enolan: I'm following the [standardized conventions](https://doc.rust-lang.org/1.0.0/style/style/naming/README.html). I don't have a strong preference either, but given that there exists a standard I'd prefer the consistency.

enolan:
I think I got everything you asked for. Added support for some of the browser-specific events as well.

src/transports/eip_1193.rs
tomusdrw: Yeah, I'm not objecting to the standard, both `eip_1993` and `eip1993` are valid module names and both adhere to the standard. It's just they come from different "source strings" `EIP-1993` and `EIP1993`. 
As I said I'm not super strong on this, if `EIP-1993` is the only correct EIP name, so bit it, I can live with `eip_1993` module name :)

src/api/accounts.rs
```diff
@@ -92,72 +47,132 @@ impl<T: Transport> Accounts<T> {
...
+                .ok_or_else(|| error::Error::Recovery(signing::RecoveryError::InvalidSignature))?;
+            let address = signing::recover(message_hash.as_bytes(), &signature, recovery_id)?;
+            Ok(address)
+        }
     }
 }
 
 /// A transaction used for RLP encoding, hashing and signing.
+#[cfg(feature = "signing")]
 struct Transaction {
```
tomusdrw, line 176: If `Transaction` is only used in case of signing, why not move it to the `mod signing` along with all the use statements? The goal is to minimize the number of `#[cfg(feature = "signing")]` :)

src/contract/mod.rs
```diff
@@ -1,12 +1,17 @@
 //! Ethereum Contract Interface
 
-use crate::api::{Accounts, Eth, Namespace};
+#[cfg(feature = "signing")]
```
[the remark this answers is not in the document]
tomusdrw: Could you please group all signing related things into a submodule here as well? Basically I'd like a single `#[cfg(feature = "signing")]` stament instead of sprinkling it all around.

src/signing.rs
```diff
@@ -1,10 +1,16 @@
...
+use crate::types::Address;
+use crate::types::H256;
+#[cfg(feature = "signing")]
 use secp256k1::recovery::{RecoverableSignature, RecoveryId};
+#[cfg(feature = "signing")]
 use secp256k1::{Message, PublicKey, Secp256k1};
+#[cfg(feature = "signing")]
 use std::ops::Deref;
 
+#[cfg(feature = "signing")]
```
tomusdrw, line 13: And here as well. Perhaps it's better to simply disable the entire `signing.rs` file?

tomusdrw (commented):
Looks really good already, I'd still like to minimize the number of feature-gated lines though (ideally just one sub-module per file).

enolan:
Done. I decided against disabling the whole `signing.rs` file, since it has the `keccak256` function which doesn't need `secp256k1` and I didn't see an obvious place to move it to.

tomusdrw (approved):
Thanks a lot!

PR #418, opened 2020-12-09
Port to hex-literal and hex

vorot93:
Better readability, stronger typing and less imports in tests.


tomusdrw (approved):
Awesome, thanks! Please `cargo fmt` to fix the build.

Issue #415, opened 2020-11-29
IPC support missing


lsunsi:
I know it was removed recently, but since I'm interested on IPC support and I was left wondering why and how could we get it back, I decided to open this issue up so we can discuss.

What do we need to get the IPC support back?
I'm willing to help if I can.

tomusdrw:
Previously the IPC support was built using [`tokio-uds`](https://crates.io/crates/tokio-uds/0.2.7), but afaict it stil hasn't migrated to new futures (and possibly it never will, cause afaict there is `UnixStream` within `tokio` directly now).

So the transport needs a complete rewrite either using [`UnixStream`](https://docs.rs/tokio/0.3.4/tokio/net/struct.UnixStream.html) (*nix only) or something like [`parity-tokio-ipc](https://github.com/NikVolf/parity-tokio-ipc) for Windows & *nix compatibility.

lsunsi:
@tomusdrw I'm currently working on a implementation using the UnixStream from Tokio for the IPC transport based on the current WebSocket and old IPC implementations. You can expect a pull request about this in a couple weeks, or maybe less.

lsunsi:
@tomusdrw Awesome! Now we're probably going to add the batch transport, or at least the duplex one (because we need it in our project). haha.

tomusdrw:
Looking forward to! :)

PR #424, opened 2020-12-15
Add IPC Transport

lsunsi:
Regarding this issue: https://github.com/tomusdrw/rust-web3/issues/415

This PR does not close it yet, because it doesn't add the duplex and batch transports.
But it's a start.

### DIFF
--- a/src/transports/ipc.rs
+++ b/src/transports/ipc.rs
@@ -1,0 +1,289 @@
+//! IPC transport
+
+use crate::{helpers, Error, RequestId, Result, Transport};
+use jsonrpc_core as rpc;
+use std::{
+    collections::BTreeMap,
+    path::Path,
+    pin::Pin,
+    sync::{atomic::AtomicUsize, Arc},
+    task::{Context, Poll},
+};
+use tokio::{
+    io::{reader_stream, AsyncWriteExt},
+    net::UnixStream,
+    stream::StreamExt,
+    sync::{mpsc, oneshot},
+};
+
+/// Unix Domain Sockets (IPC) transport.
+#[derive(Debug, Clone)]
+pub struct Ipc {
+    id: Arc<AtomicUsize>,
+    messages_tx: mpsc::UnboundedSender<TransportMessage>,
+}
+
+#[cfg(unix)]
+impl Ipc {
+    /// Creates a new IPC transport from a given path.
+    ///
+    /// IPC is only available on Unix.
+    pub async fn new<P: AsRef<Path>>(path: P) -> Result<Self> {
+        let stream = UnixStream::connect(path).await?;
+
+        Ok(Self::with_stream(stream))
+    }
+
+    fn with_stream(stream: UnixStream) -> Self {
+        let id = Arc::new(AtomicUsize::new(1));
+        let (messages_tx, messages_rx) = mpsc::unbounded_channel();
+
+        tokio::spawn(run_server(stream, messages_rx));
+
+        Ipc { id, messages_tx }
+    }
+}
+
+impl Transport for Ipc {
+    type Out = Response;
+
+    fn prepare(&self, method: &str, params: Vec<rpc::Value>) -> (crate::RequestId, rpc::Call) {
+        let id = self.id.fetch_add(1, std::sync::atomic::Ordering::AcqRel);
+        let request = helpers::build_request(id, method, params);
+        (id, request)
+    }
+
+    fn send(&self, id: RequestId, request: rpc::Call) -> Self::Out {
+        let (response_tx, response_rx) = oneshot::channel();
+
+        let message = TransportMessage {
+            request: rpc::Request::Single(request),
+            response_tx,
+            id,
+        };
+
+        Response(self.messages_tx.send(message).map(|()| response_rx).map_err(Into::into))
+    }
+}
+
+/// A future representing a pending RPC request. Resolves to a JSON RPC value.
+pub struct Response(Result<oneshot::Receiver<rpc::Value>>);
+
+impl futures::Future for Response {
+    type Output = Result<rpc::Value>;
+    fn poll(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
+        match &mut self.0 {
+            Err(err) => Poll::Ready(Err(err.clone())),
+            Ok(ref mut rx) => {
+                let value = ready!(futures::Future::poll(Pin::new(rx), cx))?;
+                Poll::Ready(Ok(value))
+            }
+        }
+    }
+}
+
+#[derive(Debug)]
+struct TransportMessage {
+    id: RequestId,
+    request: rpc::Request,
+    response_tx: oneshot::Sender<rpc::Value>,
+}
+
+#[cfg(unix)]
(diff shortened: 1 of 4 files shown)


src/transports/ipc.rs
```diff
@@ -0,0 +1,185 @@
...
+                            }
+                            read_buffer.truncate(remaining_bytes_len);
+
+                            let outputs = match response {
+                                rpc::Response::Single(output) => vec![output],
+                                rpc::Response::Batch(outputs) => outputs,
+                            };
+
+                            for output in outputs {
+                                let id = match &output {
```
tomusdrw, line 126: This could simply be `let id = output.id().clone()`?

src/transports/ipc.rs
```diff
@@ -0,0 +1,185 @@
...
+                                read_buffer[i] = read_buffer[pos + i];
+                            }
+                            read_buffer.truncate(remaining_bytes_len);
+
+                            let outputs = match response {
+                                rpc::Response::Single(output) => vec![output],
+                                rpc::Response::Batch(outputs) => outputs,
+                            };
+
+                            for output in outputs {
```
tomusdrw, line 125: I'd split the content of the loop into a function:
```rust
fn respond(pending_response_tx: &mut _, output: rpc::Output) -> Result<(), ()> {
...
}
```

Which would allow you to use `?` to early exit from the function in case of errors, you could avoid `continue` statements and I think it would make the entire `loop { select!` stuff smaller and easier to follow.
```rust
for output in outputs {
  let _ = respond(&mut pending_response_tx, output);
}
```

src/transports/ipc.rs
```diff
@@ -0,0 +1,185 @@
...
+                                        continue;
+                                    },
+                                };
+
+                                if let Err(err) = response_tx.send(value) {
+                                    log::warn!("Sending a response to deallocated channel: {:?}", err);
+                                }
+                            }
+                        }
+                        Err(err) => log::warn!("Got bad IPC response bytes: {:?}", err),
```
tomusdrw, line 160: Should it really be a `warn`? What if we simply didn't get a full response yet (but it does have `}` or `]`)?

src/transports/ipc.rs
```diff
@@ -0,0 +1,185 @@
...
+                                    rpc::Output::Success(success) => success.id.clone(),
+                                    rpc::Output::Failure(failure) => failure.id.clone(),
+                                };
+
+                                let value = match helpers::to_result_from_output(output) {
+                                    Ok(value) => value,
+                                    Err(err) => {
+                                        log::warn!("Unable to parse output into rpc::Value: {:?}", err);
+                                        continue;
+                                    },
```
tomusdrw, line 136: If you extract to a function this would become
```suggestion
                                let value = helpers::to_result_from_output(output)
                                    .map_err(|err| log::warn!("Unable to parse output into rpc::Value: {:?}", err))?;
```

which imho looks much nicer

src/transports/ipc.rs
```diff
@@ -0,0 +1,185 @@
...
+                                    log::warn!("Sending a response to deallocated channel: {:?}", err);
+                                }
+                            }
+                        }
+                        Err(err) => log::warn!("Got bad IPC response bytes: {:?}", err),
+                    };
+                },
+                Some(Err(err)) => {
+                    log::error!("IPC read error: {:?}", err);
+                    break;
```
tomusdrw, line 165: Should this be `return Err(...)` instead?

src/transports/ipc.rs
```diff
@@ -0,0 +1,185 @@
...
+                },
+                None => break,
```
tomusdrw, line 167: I'd change this to `return Ok(())` and make the whole `loop` the last statement in the function.

src/transports/mod.rs
```diff
@@ -15,6 +15,9 @@ pub mod ws;
 #[cfg(any(feature = "ws-tokio", feature = "ws-async-std"))]
 pub use self::ws::WebSocket;
 
+pub mod ipc;
```
tomusdrw, line 18: Could you feature gate this with `ipc-tokio` and make sure that features are correctly configured in `Cargo.toml` file?

Also please add a test to `.github/workflows/rust.yml` to make sure we test the transport in isolation.

src/transports/ipc.rs
```diff
@@ -0,0 +1,185 @@
...
+    let (socket_reader, mut socket_writer) = unix_stream.into_split();
+    let mut pending_response_txs = BTreeMap::default();
+
+    let mut socket_reader = reader_stream(socket_reader);
+    let mut messages_rx = messages_rx.fuse();
+    let mut read_buffer = vec![];
+
+    loop {
+        tokio::select! {
+            message = messages_rx.next() => match message {
```
tomusdrw, line 89: The `match` here could be `if let Some(message) = message {...}`

tomusdrw (commented):
Thank you, looks good! Would be nice to have some docs and tests.

lsunsi:
@tomusdrw Thanks for the review! We're finishing up with the doc and tests, along with adding all the cfg flags. Then we'll take a look at your comments (and remove the draft flag). Be right back!

src/transports/ipc.rs
FelipeRosa: Nice catch

src/transports/ipc.rs
FelipeRosa: Done in 93d190f08b744c5d484adcd4c0e03f37c1e40703

src/transports/ipc.rs
FelipeRosa: Done in 7f34638c896677321818bb0c1d5a93a2ad7ab3f4

src/transports/ipc.rs
FelipeRosa: Done in 6a28cade6a92ba553f77afdfac95860a4873b5ca

src/transports/ipc.rs
FelipeRosa: Done in 7f34638c896677321818bb0c1d5a93a2ad7ab3f4

src/transports/ipc.rs
FelipeRosa: Done in e3d3469b9fe9de4ff105267b1c741a8919b2dc65

src/transports/ipc.rs
FelipeRosa: Done in e3d3469b9fe9de4ff105267b1c741a8919b2dc65

src/transports/mod.rs
FelipeRosa: The changes to the `.github/workflows/rust.yml` file are in 583d81638ac431cc75d7b77b05578171c061ffb5.

The feature gate was amended to the first commit.

src/transports/ipc.rs
FelipeRosa: Done in 1788be6f69e1a7864572b35f1c396091cfdfd014

lsunsi:
@tomusdrw Done deal! All comments replied. We sent fixups in order to make your re-review easier, but we'll squash all commits into one if everything is fine. Just let us know.

tomusdrw (approved):
Looks perfect, thank you so much! I squash PR commits before merge anyway, so need to squash manually.

PR #428, opened 2020-12-17
Implement DuplexTransport for IPC

### LINKED ISSUE #426: IPC duplex transport support
Follow up on #415. Add `DuplexTransport` implementation for IPC.

lsunsi:
Closes #426

### DIFF
--- a/src/transports/ipc.rs
+++ b/src/transports/ipc.rs
@@ -1,6 +1,6 @@
 //! IPC transport
 
-use crate::{helpers, BatchTransport, Error, RequestId, Result, Transport};
+use crate::{api::SubscriptionId, helpers, BatchTransport, DuplexTransport, Error, RequestId, Result, Transport};
 use futures::future::{join_all, JoinAll};
 use jsonrpc_core as rpc;
 use std::{
@@ -89,6 +89,22 @@ impl BatchTransport for Ipc {
     }
 }
 
+impl DuplexTransport for Ipc {
+    type NotificationStream = mpsc::UnboundedReceiver<rpc::Value>;
+
+    fn subscribe(&self, id: SubscriptionId) -> Result<Self::NotificationStream> {
+        let (tx, rx) = mpsc::unbounded_channel();
+        self.messages_tx.send(TransportMessage::Subscribe(id, tx))?;
+        Ok(rx)
+    }
+
+    fn unsubscribe(&self, id: SubscriptionId) -> Result<()> {
+        self.messages_tx
+            .send(TransportMessage::Unsubscribe(id))
+            .map_err(Into::into)
+    }
+}
+
 /// A future representing a pending RPC request. Resolves to a JSON RPC value.
 pub struct SingleResponse(Result<oneshot::Receiver<rpc::Value>>);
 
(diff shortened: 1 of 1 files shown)


src/transports/ipc.rs
```diff
@@ -218,31 +246,65 @@ async fn run_server(unix_stream: UnixStream, messages_rx: mpsc::UnboundedReceive
     }
 }
 
+fn notify(
+    subscription_txs: &mut BTreeMap<SubscriptionId, mpsc::UnboundedSender<rpc::Value>>,
+    notification: rpc::Notification,
+) -> std::result::Result<(), ()> {
+    if let rpc::Params::Map(params) = notification.params {
```
tomusdrw, line 253: I feel there is a lot of overlap with `ws.rs`, but I'm okay to abstract it away in case we get the third duplex transport.

src/transports/ipc.rs
```diff
@@ -218,31 +246,65 @@ async fn run_server(unix_stream: UnixStream, messages_rx: mpsc::UnboundedReceive
...
-    let id = output.id().clone();
+    let outputs = match response {
+        rpc::Response::Single(output) => vec![output],
+        rpc::Response::Batch(outputs) => outputs,
+    };
 
-    let value = helpers::to_result_from_output(output).map_err(|err| {
-        log::warn!("Unable to parse output into rpc::Value: {:?}", err);
-    })?;
+    for output in outputs {
```
tomusdrw, line 283: This changes semantics of batch responses quite significantly actually.
Previously in case of error with single `output` you'd continue with others - now we will early exit the function.
Please consider extracting the inner-part of the loop to a scoped function or something, so that we conitnue with other outputs in case of a failure.
(might be good to add a test for it too :))

src/transports/ipc.rs
lsunsi: I agree with you on that one, but we were focused on bringing the duplex support back without messing around with stuff too much haha. Is it cool to leave it like that for now?

src/transports/ipc.rs
lsunsi: You are right! We didn't realise this at the time, but definitely going to fix (and test) it. Brb

src/transports/ipc.rs
tomusdrw: Yup!

src/transports/ipc.rs
lsunsi: @tomusdrw Just amended the fix and added a test that would break had the fix not been made!

src/transports/ipc.rs
```diff
@@ -183,30 +212,29 @@ async fn run_server(unix_stream: UnixStream, messages_rx: mpsc::UnboundedReceive
                 Some(Ok(bytes)) => {
                     read_buffer.extend_from_slice(&bytes);
 
-                    let pos = match read_buffer.iter().rposition(|&b| b == b']' || b == b'}') {
-                        Some(pos) => pos + 1,
-                        None => continue,
-                    };
+                    let read_len = {
+                        let mut de: serde_json::StreamDeserializer<_, serde_json::Value> =
```
tomusdrw, line 216: That's awesome! I hope it works in case there is no `\n` between requests, right? I remember this being the biggest issue, but reading the documentation it seems it does so I'm super happy to get rid of the hacky code we had previously.

tomusdrw (approved):

tomusdrw:
Thanks a lot @lsunsi !!!

PR #429, opened 2020-12-17
Fix IPC server shutdown

lsunsi:
The problem was that the server might stay alive forever
due to the UnixSocket being open. What this change does is:

If there is no more requests (the requests channel closed),
then we just keep the server open if there are subscribers
or pending requests. Otherwise it is considered drained and
it closes itself to avoid leakage.

### DIFF
--- a/src/transports/ipc.rs
+++ b/src/transports/ipc.rs
@@ -150,60 +150,60 @@ enum TransportMessage {
 }
 
 #[cfg(unix)]
-async fn run_server(unix_stream: UnixStream, messages_rx: mpsc::UnboundedReceiver<TransportMessage>) -> Result<()> {
-    let (socket_reader, mut socket_writer) = unix_stream.into_split();
+async fn run_server(mut unix_stream: UnixStream, messages_rx: mpsc::UnboundedReceiver<TransportMessage>) -> Result<()> {
+    let (socket_reader, mut socket_writer) = unix_stream.split();
     let mut pending_response_txs = BTreeMap::default();
     let mut subscription_txs = BTreeMap::default();
 
     let mut socket_reader = reader_stream(socket_reader);
     let mut messages_rx = messages_rx.fuse();
     let mut read_buffer = vec![];
+    let mut closed = false;
 
-    loop {
+    while !closed || pending_response_txs.len() > 0 {
         tokio::select! {
-            message = messages_rx.next() => if let Some(message) = message {
-                match message {
-                    TransportMessage::Subscribe(id, tx) => {
-                        if let Some(_) = subscription_txs.insert(id.clone(), tx) {
-                            log::warn!("Replacing a subscription with id {:?}", id);
-                        }
-                    },
-                    TransportMessage::Unsubscribe(id) => {
-                        if let None = subscription_txs.remove(&id) {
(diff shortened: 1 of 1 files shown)


lsunsi:
@tomusdrw This is the last PR I can think of for now. While writing the duplex I realised the server could outlive it's usefulness and leak. Not anymore after change, which guarantees that it dies when requests, pending responses and subscriptions are no more.

src/transports/ipc.rs
```diff
@@ -158,52 +158,52 @@ async fn run_server(unix_stream: UnixStream, messages_rx: mpsc::UnboundedReceive
     let mut socket_reader = reader_stream(socket_reader);
     let mut messages_rx = messages_rx.fuse();
     let mut read_buffer = vec![];
+    let mut closed = false;
 
-    loop {
+    while !closed || pending_response_txs.len() > 0 || subscription_txs.len() > 0 {
```
tomusdrw, line 163: I feel this would be blocking the server anyway, since I think we don't really automatically unsubscribe, no?

I'm not entirely sure what the right behaviour should be though, but I feel that being able to close the client upon drop should disregard all subscriptions that are still open, we should wait for any incomplete requests though. I'm not 100% sure about this, interested about your motivation on this.

```suggestion
    while !closed || pending_response_txs.len() > 0 {
```

lsunsi: I agree with you entirely, it'd be best not to count the subscriptions since they're not automatic. That said, thinking about this, me and @FelipeRosa thought it'd be cool to rely on the channel state of subscriptions in order to clean up automatically the subscriptions instead of error-ing it out.

We added the 4593a2f commit that implements this: if a subscription channel is closed, it automatically unsubscribes, which means the check against open subscriptions would be better. In case the user of the client drops the clients but keep the receivers of subscriptions, server remains alive for as long as it holds the receiver.

PS: There's a flaw on this, which is to say that the auto cleaning just happens upon getting a message. We realised this, but thought it was a small trade off for having all the other cases work accordingly.

What do you think?

lsunsi:
@tomusdrw If is there anything else you can let me know! (:

tomusdrw:
@lsunsi Apologies for the delay, I'll take a closer look today evening.

lsunsi:
@tomusdrw no worries (: thanks for the attention. I just updated the branch with the newest master version.

src/transports/ipc.rs
tomusdrw: First of all, apologies for the delay! I've been putting this off for way too long. I don't think this is a good solution though, there might be subscriptions that fire only very rarely and this would make closing quite unpredictable.

Why is receiving subscriptions important in case the server is closing? The nature of subscriptions should be ephemeral anyway - i.e. they receive a bunch of notifications, but there are no super strict guarantees about them. 

IMHO if the server is closing up, we should just complete any requests that are still pending and just finish up, despite subscriptions being active, but I'm happy to hear your arguments :)

lsunsi: @tomusdrw No arguments here, I agree with you. Both are valid approaches, so I changed the branch to reflect what you suggested. (:

tomusdrw (approved):

PR #449, opened 2021-01-13
add method to import private key to node

justindthomas:
I treat my Ethereum Node as ephemeral (private chain, using a dedicated node for sealing and Kubernetes members to process transactions). Private keys are stored in a HashiCorp Vault to pull in to the ephemeral Ethereum Node when needed.

That written, I need access to the private keys to store in HCV. This snippet just exposes the `importRawKey` method on the Ethereum node to an application via `web3`.

I'm a novice Rust user and had some trouble getting the `private_key` in a format usable for this call. I initially had this method accept an `H256`, but couldn't figure out how to easily output the contents without the `0x` so that Geth would accept it. I settled on using `&hex::encode(&bytes)` to generate the argument as an `&str` for the method here.

### DIFF
--- a/src/api/personal.rs
+++ b/src/api/personal.rs
@@ -81,7 +81,7 @@ impl<T: Transport> Personal<T> {
     /// Imports a raw key and protects it with the given password.
     /// Returns the address of created account.
     pub fn import_raw_key(&self, private_key: &[u8; 32], password: &str) -> CallFuture<Address, T::Out> {
-      let private_key = hex::encode(private_key);
+        let private_key = hex::encode(private_key);
         let private_key = helpers::serialize(&private_key);
         let password = helpers::serialize(&password);
 
@@ -166,7 +166,7 @@ mod tests {
     );
 
     rpc_test! {
-      Personal:import_raw_key, "0000000000000000000000000000000000000000000000000000000000000123", "hunter2" =>
+      Personal:import_raw_key, &[0u8; 32], "hunter2" =>
       "personal_importRawKey", vec![r#""0000000000000000000000000000000000000000000000000000000000000123""#, r#""hunter2""#];
       Value::String("0x0000000000000000000000000000000000000123".into()) => Address::from_low_u64_be(0x123)
     }
--- a/src/api/personal.rs
+++ b/src/api/personal.rs
@@ -163,4 +163,10 @@ mod tests {
       ::serde_json::from_str(EXAMPLE_TX).unwrap()
       => ::serde_json::from_str::<RawTransaction>(EXAMPLE_TX).unwrap()
     );
+
+    rpc_test! {
+      Personal:import_private_key, "0000000000000000000000000000000000000000000000000000000000000123", "hunter2" =>
+      "personal_importRawKey", vec![r#""0000000000000000000000000000000000000000000000000000000000000123""#, r#""hunter2""#];
+      Value::String("0x0000000000000000000000000000000000000123".into()) => Address::from_low_u64_be(0x123)
+    }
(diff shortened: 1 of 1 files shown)


justindthomas:
I have literally no idea what I'm doing with that test. I tried to guess at what I could glean from the documentation in `helpers.rs` and the previous examples. It passes, but ¯\_(ツ)_/¯

justindthomas:
Here's an example of how this method is used.

`````
let mut secret = [0u8; 32];
thread_rng().fill_bytes(&mut secret);
let public_key = web3.personal().import_raw_key(&hex::encode(&secret), "hunter2").await.unwrap()
`````

File: src/api/personal.rs [hunk 1]
```diff
@@ -77,6 +77,18 @@ impl<T: Transport> Personal<T> {
                 .execute("personal_signTransaction", vec![transaction, password]),
         )
     }
+
+    /// Imports a raw key and protects it with the given password.
+    /// Returns the address of created account.
+    pub fn import_raw_key(&self, private_key: &str, password: &str) -> CallFuture<Address, T::Out> {
```
tomusdrw, line 83:
```suggestion
    pub fn import_raw_key(&self, private_key: &[u8; 32], password: &str) -> CallFuture<Address, T::Out> {
      let private_key = hex::encode(private_key);
```
If the private key is expected to be hexadecimal, let's enforce that on the API level via typesystem.

I think using `U256` and then `hex::encode` should work too.

src/api/personal.rs
```diff
@@ -151,4 +163,10 @@ mod tests {
       ::serde_json::from_str(EXAMPLE_TX).unwrap()
       => ::serde_json::from_str::<RawTransaction>(EXAMPLE_TX).unwrap()
     );
+
+    rpc_test! {
+      Personal:import_raw_key, "0000000000000000000000000000000000000000000000000000000000000123", "hunter2" =>
```
tomusdrw, line 168: What the test is doing is calling `Personal::import_raw_key` with the two parameters you have here, next line checks that it's properly serialized as JSON-RPC request and the last line is JSON-RPC response and how it's supposed to be deserialized. So it looks good then :)

tomusdrw (commented):
Looks pretty good! I'd just make the API more type safe by enforcing the private key to be `&[u8; 32]` instead of a `str`.

src/api/personal.rs
justindthomas: Great; thank you!

src/api/personal.rs
```diff
@@ -81,7 +81,7 @@ impl<T: Transport> Personal<T> {
     /// Imports a raw key and protects it with the given password.
     /// Returns the address of created account.
     pub fn import_raw_key(&self, private_key: &[u8; 32], password: &str) -> CallFuture<Address, T::Out> {
-      let private_key = hex::encode(private_key);
+        let private_key = hex::encode(private_key);
         let private_key = helpers::serialize(&private_key);
```
justindthomas, line 85: Is it idiomatically acceptable to shadow variables repeatedly like this? I'm fine with it, just curious about the general community expectations.

justindthomas:
I merged the updated commits to master in to this branch, but it looks like there's some trouble with the recent parity changes.

src/api/personal.rs
tomusdrw: Yes, they are actually not shadowed, but rather re-declared (previous one can be dropped) and afaik it's totally acceptable pattern.

tomusdrw (approved):

tomusdrw:
Thanks!

PR #450, opened 2021-01-16
remove explicit native-tls dependency

toxeus:
`native-tls` is actually only an implicit dependency which might change in the future.


tomusdrw (approved):

tomusdrw:
Thanks!

Issue #440, opened 2021-01-03
support for `eth_pendingTransactions`


mohoff:
Hi, it seems that the call `eth_pendingTransactions` is currently not supported, while getting a pending block is supported via `block(BlockNumber::Pending.into)` as #404 claims. [A stackoverflow answer ](https://ethereum.stackexchange.com/a/6725) suggests that `eth_pendingTransactions` is Geth specific and thus maybe the reason why it isn't supported here?

If mistaken, I can give the implementation a try with some hints.

tomusdrw:
hey! Happy to accept the PR, the library was originally based on Parity's/OpenEthereum's RPCs, but tries not to be biased towards any particular client (if it's within the RPC spec).

mohoff:
Thanks @tomusdrw . I researched a bit more and found that getting pending transactions is not part of the [standard JSON RPC API](https://eth.wiki/json-rpc/API) but can be found under different names in client-specific calls:

- geth: [`txpool_content`](https://geth.ethereum.org/docs/rpc/ns-txpool#txpool_content). Already supported [in `api/tx_pool.rs`](https://github.com/tomusdrw/rust-web3/blob/b9028631a159c5932f09ecebbf36560481a1903b/src/api/txpool.rs#L32).
- parity/openethereum: [`parity_pendingtransactions`](https://openethereum.github.io/JSONRPC-parity-module#parity_pendingtransactions). Currently not supported although several other `parity_*` calls do exist in this crate.

What is confusing for me is that [`web3.js` supports a `eth_pendingTransactions` call](https://github.com/ethereum/web3.js/blob/4469516316d55e35c5ed43dcb883bcf34a6f337c/packages/web3-eth/src/index.js#L557) but implements none of the official client-specific calls listed above. What do I fail to understand here?

Assuming my digging is correct, we can just add support for `parity_pendingTransactions` and call it a day. This would enable users to fetch pending transactions for both clients. Additionally, does a client-agnostic `eth().get_pending_transactions()` make sense for you as part of this crate?

tomusdrw:
> What is confusing for me is that web3.js supports a eth_pendingTransactions call but implements none of the official client-specific calls listed above. What do I fail to understand here?

Might be some historical reasons, or for instance compatibility with Ganache?

> Assuming my digging is correct, we can just add support for parity_pendingTransactions and call it a day. This would enable users to fetch pending transactions for both clients.

Yes, I like the idea.

[...]

PR #451, opened 2021-01-19
feat: support parity_pendingTransactions

mohoff:
Adding support for [`parity_pendingTransactions`](https://openethereum.github.io/JSONRPC-parity-module#parity_pendingtransactions). Implements https://github.com/tomusdrw/rust-web3/issues/440

Applying a pending transaction filter is blocked by https://github.com/openethereum/openethereum/issues/159. However, omitting a filter gives a full list of pending transactions as expected. This has been tested with a synced openethereum node.

Would love to get some feedback!

### DIFF
--- a/src/api/parity.rs
+++ b/src/api/parity.rs
@@ -40,10 +40,11 @@ impl<T: Transport> Parity<T> {
         limit: Option<usize>,
         filter: Option<ParityPendingTransactionFilter>,
     ) -> CallFuture<Vec<Transaction>, T::Out> {
+        let limit = limit.map(Value::from);
+        let filter = filter.as_ref().map(helpers::serialize);
         let params = match (limit, filter) {
-            (Some(l), Some(f)) => vec![l.into(), helpers::serialize(&f)],
-            (Some(l), _) => vec![l.into()],
-            (_, Some(f)) => vec![Value::Null, helpers::serialize(&f)],
+            (l, Some(f)) => vec![l.unwrap_or(Value::Null), f],
+            (Some(l), None) => vec![l],
             _ => vec![],
         };
 
@@ -57,7 +58,7 @@ mod tests {
     use crate::{
         api::Namespace,
         rpc::Value,
-        types::{Address, CallRequest, FilterCondition, ParityPendingTransactionFilterBuilder, Transaction, U64},
+        types::{Address, CallRequest, FilterCondition, ParityPendingTransactionFilter, Transaction, U64},
     };
     use hex_literal::hex;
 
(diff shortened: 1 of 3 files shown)


src/types/parity_pending_transaction.rs
```diff
@@ -0,0 +1,135 @@
...
+where
+    T: Serialize,
+{
+    fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: Serializer,
+    {
+        let mut map = serializer.serialize_map(Some(1))?;
+        match self {
+            Self::Lt(v) => map.serialize_entry("lt", v),
```
tomusdrw, line 61: I'm pretty sure it's totally possible to do with `serde` attributes instead of writing a custom serialize implementation.

src/types/parity_pending_transaction.rs
```diff
@@ -0,0 +1,135 @@
...
+    Serialize,
+};
+
+use super::{Address, U256, U64};
+
+/// Condition to filter pending transactions
+#[derive(Clone)]
+pub enum FilterCondition<T> {
+    /// Lower Than
+    Lt(T),
```
tomusdrw, line 12: I'd name enum variants with full names and only shorten it for serialisation.

src/types/parity_pending_transaction.rs
```diff
@@ -0,0 +1,135 @@
...
+    }
+
+    /// Sets `to`
+    pub fn to(mut self, to_or_action: ToFilter) -> Self {
+        self.filter.to = Some(to_or_action);
+        self
+    }
+
+    /// Sets `gas`
+    pub fn gas(mut self, gas: FilterCondition<U64>) -> Self {
```
tomusdrw, line 108: For convenience you might do this:
```rust
pub fn gas(mut self, gas: impl Into<FilterCondition<U64>>) -> Self {
...
}
```
and implement:
```rust
impl From<T> for FilterCondition<T> {
 fn from(t: T) -> Self {
   FilterCondition::Equal(t)
 }
}
```

This will allow you to use the API without necessarily wrapping with `FilterCondition` in case what you want is exact values (which I believe is the most common use case anyway):
```
PendingTransactionFilter::builder()
  .gas(U64::from(5))
  .build()
```

src/types/parity_pending_transaction.rs
```diff
@@ -0,0 +1,135 @@
...
+pub enum ToFilter {
+    /// Address
+    Address(Address),
+    /// Action (i.e. contract creation)
+    Action,
+}
+
+/// Filter for pending transactions (only openethereum/Parity)
+#[derive(Clone, Default, Serialize)]
+pub struct ParityPendingTransactionFilter {
```
tomusdrw, line 30: Could you please add `::builder()` function to make it easy to create builder struct without importing the type?
```rust
impl ParityPendingTransactionFilter {
  pub fn builder() -> ParityPendingTransactionFilterBuilder {
    Default::default()
  }  
}
```

src/api/parity.rs
```diff
@@ -31,6 +32,23 @@ impl<T: Transport> Parity<T> {
...
+        &self,
+        limit: Option<usize>,
+        filter: Option<ParityPendingTransactionFilter>,
+    ) -> CallFuture<Vec<Transaction>, T::Out> {
+        let params = match (limit, filter) {
+            (Some(l), Some(f)) => vec![l.into(), helpers::serialize(&f)],
+            (Some(l), _) => vec![l.into()],
+            (_, Some(f)) => vec![Value::Null, helpers::serialize(&f)],
+            _ => vec![],
+        };
```
tomusdrw, line 48:
```suggestion
     let limit = limit.map(Value::from);
     let filter = filter.as_ref().map(helpers::serialize);
     let params = match (limit, filter) {
            (l, Some(f)) => vec![l.unwrap_or(Value::Null), f],
            (Some(l), None) => vec![l],
            _ => vec![],
        };
```
This avoids duplication.

tomusdrw (commented):
looks good! I've left couple of improvements suggestions, let me know wdyt.

mohoff:
@tomusdrw Addressed all feedback and learnt a few things, thank you!

I also updated the test to use the example transaction from the [parity docs](https://openethereum.github.io/JSONRPC-parity-module#parity_pendingtransactions)

src/types/parity_pending_transaction.rs
mohoff: Cool trick!

tomusdrw (approved):
looks awesome, thanks!

Issue #421, opened 2020-12-12
Async web3::contract::Contract::query for Batch Transports


mdben1247:
Hi, before `v0.14.0` one could call something like `let _ = contract.query(...);` for contracts that had batch web3 transport to queue the request that was later sent with `submit_batch()`. But, if I understand correctly, the new async version of `web3::contract::Contract::query` wouldn't do anything unless it was called with `web3::block_on`. And even then, the return value of the query function is `Result<R: Detokenize>`, and it doesn't seem clear what that should be in such a case, since the actual response is received only later when submitting the batch.

mdben1247:
I did some testing and indeed, using batch transport with contracts seems to be broken in `v0.14.0`. The issue is that all contract functions are now `async`, so they don't get executed at all to push the request on batch list. If you block on them then you have another problem: you can't execute them later to collect results after submitting the batch. The workaround is to rewrite the functions to be only partially `async`, so the first part gets executed - `eth.call()` will push the request on batch list - and only then the future is returned. For `web3::contract::Contract::query` this workaround works for me:

```
    /// Call constant function
    pub fn query<R, A, B, P>(&self, func: &str, params: P, from: A, options: Options, block: B) -> impl Future<Output = Result<R>> + '_
    where
        R: Detokenize,
        A: Into<Option<Address>>,
        B: Into<Option<BlockId>>,
        P: Tokenize,
    {

        let result = self.abi
            .function(func)
            .and_then(|function| {
                function
                    .encode_input(&params.into_tokens())
                    .map(|call| (call, function))
            })
            .map(|(call, function)| {
                let call_future = self.eth.call(
                    CallRequest {
                        from: from.into(),
                        to: Some(self.address),
                        gas: options.gas,
                        gas_price: options.gas_price,
                        value: options.value,
                        data: Some(Bytes(call)),
                    },
                    block.into(),
                );
                (call_future, function)
            });

        async {
            let (call_future, function) = result?;
            let bytes = call_future.await?;
            let output = function.decode_output(&bytes.0)?;
            R::from_tokens(output)
        }
    }
```

This fix keeps the function signature unchanged, but ideally the returned future would be wrapped into result, so any ethabi errors regarding function name and encoding input would be returned right away, not as part of the future.

tomusdrw:
[...]

PR #453, opened 2021-01-21
batch transport contrat query fix

mdben1247:
This is a fix for #421. It only fixes the `query` function, other async functions (`call`, `esitmate_gas`, `events`) remain broken with batch transport. We only use `query` in our application, so I can confirm it's working well. A bit afraid to mess with the others without a clear way to test it.

### DIFF
--- a/src/contract/mod.rs
+++ b/src/contract/mod.rs
@@ -203,14 +203,22 @@ impl<T: Transport> Contract<T> {
     }
 
     /// Call constant function
-    pub fn query<R, A, B, P>(&self, func: &str, params: P, from: A, options: Options, block: B) -> impl Future<Output = Result<R>> + '_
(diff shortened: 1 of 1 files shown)


tomusdrw:
@mdben1247 thanks a lot! Unfortunately I feel this is just a band aid and the batch transport has to be redesigned completely, I'm pretty sure all other functions don't work correctly either.

mdben1247:
Yes, it's a band aid. It's also very weird, because the non-async part can actually produce an error, but we don't return that error right away, only latter when await-ing on the async-part. If we wanted to fix that another `Result<>` would be needed in the return value of the function, but that would be very inconvenient and unnecessary for ordinary (non-batch transport) use case.

tomusdrw (approved):

PR #462, opened 2021-02-23
workaround for issue in serde_json with arbitrary_precision feature

### LINKED ISSUE #460: serde deserialization fails if arbitrary-precision is enabled
If a user is using the `"arbitrary-precision"` feature of `serde`, the deserialization occurring in the following line fails:

https://github.com/tomusdrw/rust-web3/blob/11277e07fd4ebbb4209a96860cb00a0184115016/src/transports/http.rs#L193

Error returned:
`Error("data did not match any variant of untagged enum Response", line: 0, column: 0)`

Workaround:
If the user has arbitrary-precision activated an extra step must be done:
```
fn single_response<T: Deref<Target = [u8]> + std::fmt::Debug>(response: T) -> error::Result<rpc::Value> {
    let value = serde_json::from_slice::<rpc::Value>(&*response).map_err(|e| Error::InvalidResponse(format!("Parse error: {:?}", e)))?;
    let response = serde_json::from_value(value).map_err(|e| Error::InvalidResponse(format!("Parse error: {:?}", e)))?;

    match response {
        rpc::Response::Single(output) => helpers::to_result_from_output(output),
        _ => Err(Error::InvalidResponse("Expected single, got batch.".into())),
    }
}
```

Related issues:
- jsonrpc-client: https://github.com/paritytech/jsonrpc/issues/520
- serde-json: https://github.com/serde-rs/json/issues/505

marcbaque:
Closes #460.

It includes:
- Implementation of the workaround proposed here: https://github.com/serde-rs/json/issues/505
- New arbitrary_precision feature to enable the workaround
- Updated the CI workflow to test with the arbitrary_precision feature

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -41,6 +41,12 @@ jobs:
           command:                 test
           toolchain:               stable
           args:                    --all  --verbose
+      - name:                      Testing rust-stable with arbitrary_precision
+        uses:                      actions-rs/cargo@master
+        with:
+          command:                 test
+          toolchain:               stable
+          args:                    --all  --verbose --features arbitrary_precision
 
 ## Build Stage
       - name:                      Building rust-stable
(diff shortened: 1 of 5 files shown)


tomusdrw (approved):
Awesome, thanks!

PR #464, opened 2021-02-25
Allow tokenization of tuples

e00E:
With solidity abi v2 it has become possible to pass structs directly as
arguments to functions. A struct containing types T0 and T1 is encoded
as a tuple (T0, T1).

In the current state of rust-web3 it is not possible to pass tuples
because of the way the tokenization related traits are set up. We have
Tokenizable which is implemented for basic types like i32 and for arrays
of basic types but not for tuples of Tokenizable. Tokenize is
implemented on tuples of Tokenizable but not tuples of Tokenize so
nesting of tuples is not possible.

This PR allows tokenization of tuple types by implementing Tokenizable
on tuples of Tokenizable. In order for this to not introduce trait
ambiguity (which the compiler also notices) we have to remove the
implementations of Detokenize and Tokenize on a single T: Tokenizable.
This would be ambiguous because it would not be clear whether a (t0, t1)
refers to two function arguments or a single tuple function argument.

This last part means this is a (small) breaking change as function
arguments and return types (for example in Contract::query) now always
have to be tuples (or Vec<Token>) and cannot be single types anymore.

It might be possible to avoid the breakage by doing more trait magic to
still allow non tupled types in those places. They would implicitly be
treated as a tuple of a single element. I feel in the long term this is
more confusing as users would have to remember the special case instead
of always tupling. For example, a function taking a solidity bool would
be callable with a rust bool or (bool,) but a function taking a (bool,)
would only be callable with ((bool,),).

Note that the current release of ethabi does no handle abi v2 but there
are PRs and forks that do which ethabi will hopefully integrate.

---

Edit:
For the function return type we could do something smarter so that the tuple for single types isn't needed because function return types do not have the ambiguity as they are always a single token. This would be something we assert on top of ethabi where Function::Output is `Vec<Param>` instead of just Param for some reason. Or can abi have several output tokens against my expectation? (edit: looks like when returning a tuple older versions of solc would output a list of types instead of a single tuple type) The implementation would be as easy as using Tokenizable instead of Detokenize for the return type.

### DIFF
--- a/examples/contract.rs
+++ b/examples/contract.rs
@@ -28,8 +28,10 @@ async fn main() -> web3::contract::Result<()> {
         )
         .await?;
 
-    let result = contract.query("balanceOf", (my_account,), None, Options::default(), None);
-    let balance_of: U256 = result.await?;
+    let result: (U256,) = contract
+        .query("balanceOf", (my_account,), None, Options::default(), None)
+        .await?;
+    let balance_of = result.0;
     assert_eq!(balance_of, 1_000_000.into());
 
     // Accessing existing contract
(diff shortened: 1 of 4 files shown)


tomusdrw:
To simplify single-element return case, can we maybe introduce another variant of `query` which returns a tuple (like `query_n(`)? Or perhaps always return some kind of `QueryResult` type, which can later on be resolved to either a single value or a tuple/list?

wdyt?

e00E:
Closing this because I won't have time to work on it more. We have something similar in ethcontract https://github.com/gnosis/ethcontract-rs/blob/main/ethcontract/src/tokens.rs now.

PR #473, opened 2021-03-15
Allow signing deployment transactions with a private key

smessmer:
All previous API points to deploy contracts relied on the ethereum node to have the private keys for the account to deploy from. This PR adds an API for deploying contracts while keeping the private keys away from the ethereum node.

### DIFF
--- a/src/contract/deploy.rs
+++ b/src/contract/deploy.rs
@@ -102,6 +102,13 @@ impl<T: Transport> Builder<T> {
     /// caller to pass in a private key to sign the transaction with
     /// and therefore allows deploying from an account that the
     /// ethereum node doesn't need to know the private key for.
+    ///
+    /// An optional `chain_id` parameter can be passed to provide
+    /// replay protection for transaction signatures. Passing `None`
+    /// would create a transaction WITHOUT replay protection and
+    /// should be avoided.
+    /// You can obtain `chain_id` of the network you are connected
+    /// to using `web3.eth().chain_id()` method.
     #[cfg(feature = "signing")]
     pub async fn sign_with_key_and_execute<P, V, K>(
         self,
--- a/src/contract/deploy.rs
+++ b/src/contract/deploy.rs
@@ -1,11 +1,13 @@
 //! Contract deployment utilities
 
+#[cfg(feature = "signing")]
+use crate::signing::Key;
 use crate::{
     api::{Eth, Namespace},
     confirm,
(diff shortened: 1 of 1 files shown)


src/contract/deploy.rs
```diff
@@ -94,6 +96,49 @@ impl<T: Transport> Builder<T> {
...
+    {
+        let transport = self.eth.transport().clone();
+        let poll_interval = self.poll_interval;
+        let confirmations = self.confirmations;
+
+        self.do_execute(code, params, from.address(), move |tx| async move {
+            let tx = TransactionParameters {
+                nonce: tx.nonce,
+                to: tx.to,
+                gas: tx.gas.unwrap_or(0.into()),
```
tomusdrw, line 120: Shouldn't this be some sane default instead? Now if you don't provide `gas` the transaction will simply fail. AFAIR the nodes used to set around 1M gas for such unspecified transactions.

src/contract/deploy.rs
```diff
@@ -94,6 +96,49 @@ impl<T: Transport> Builder<T> {
...
+                gas_price: tx.gas_price,
+                value: tx.value.unwrap_or(0.into()),
+                data: tx
+                    .data
+                    .expect("Tried to deploy a contract but transaction data wasn't set"),
+                chain_id: None,
```
tomusdrw, line 126: It would be best to actually add `chain_id` here for replay protection.

tomusdrw (commented):
lgtm, except for the `gas` and `chain_id` comments.

src/contract/deploy.rs
```diff
@@ -94,6 +96,55 @@ impl<T: Transport> Builder<T> {
         .await
     }
 
+    /// Execute deployment passing code and constructor parameters.
+    ///
+    /// Unlike the above `sign_and_execute`, this method allows the
+    /// caller to pass in a private key to sign the transaction with
+    /// and therefore allows deploying from an account that the
+    /// ethereum node doesn't need to know the private key for.
```
tomusdrw, line 104:
```suggestion
    /// ethereum node doesn't need to know the private key for.
    ///
    /// An optional `chain_id` parameter can be passed to provide
    /// replay protection for transaction signatures. Passing `None`
    /// would create a transaction WITHOUT replay protection and
    /// should be avoided.
    /// You can obtain `chain_id` of the network you are connected
    /// to using `web3.eth().chain_id()` method.
```

tomusdrw (approved):
Great, thanks!

PR #484, opened 2021-04-21
Treat null miner as default in block

### LINKED ISSUE #483: rust-web3 cannot decode Infura getBlockByNumber response for pending block
because Infura sets `miner` to null.

rust-web3 has a non optional author/miner field that defaults to the 0 address if the field is not present in the json:
https://github.com/tomusdrw/rust-web3/blob/5520ec249a011ea2e5de64417054bc3d2405db27/src/types/block.rs#L18

However when testing with Infura I found that they set `miner` to `null` which makes deserialization in rust-web3 fail. Potential solutions:

* make the rust field optional which would automatically handle missing and null
* see if there is a serde attribute (maybe from the serde_with crate) that also turns null into default
* do not make any change because Infura is returning a response that is invalid according to the ethrpc specification

e00E:
When querying an Infura node for the pending block the `miner` field is
set to `null`. This used to result in a deserialization error but now
becomes the default H160 like when the field is missing.

Fixes #483

### DIFF
--- a/src/types/block.rs
+++ b/src/types/block.rs
@@ -1,5 +1,5 @@
 use crate::types::{Bytes, H160, H2048, H256, H64, U256, U64};
-use serde::{ser::SerializeStruct, Deserialize, Serialize, Serializer};
+use serde::{ser::SerializeStruct, Deserialize, Deserializer, Serialize, Serializer};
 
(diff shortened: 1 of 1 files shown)


tomusdrw (approved):
Awesome, thanks!

PR #489, opened 2021-04-30
AccessLists

mdben1247:
A rather no-frills but functional addition of AccessLists. When using access lists, transaction type must be 1. Also see: https://eips.ethereum.org/EIPS/eip-2930


tomusdrw (approved):
lgtm, thanks!

PR #491, opened 2021-05-12
switch to reqwest

### LINKED ISSUE #303: Use reqwest as client library for WASM browser support
Having a quick glance to the dependencies I see you depend on hyper directly, would it be a big effort to use reqwest instead? is one of the most used http clients this days and provides a nicer API, on top of that they already went through the effort of making it work with web assembly which seems like a major effort that this project doesn't look very motivated to take. Reqwest would act as the abstraction layer that makes the switch from a hyper backend to browser fetch at compile time.  
Nicer http api plus wasm support, sounds like a nice deal right?

clouds56:
fixes #303

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "web3"
-version = "0.15.0"
+version = "0.16.0"
 description = "Ethereum JSON-RPC client."
 homepage = "https://github.com/tomusdrw/rust-web3"
 repository = "https://github.com/tomusdrw/rust-web3"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -31,10 +31,8 @@ pin-project = "1.0"
 # Optional deps
 ## HTTP
 base64 = { version = "0.13", optional = true }
-hyper = { version = "0.14", optional = true, default-features = false, features = ["client", "http1", "stream", "tcp"] }
-hyper-tls = { version = "0.5", optional = true }
-hyper-proxy = {version = "0.9.0", default-features = false, optional = true }
-hyper-rustls = { version = "0.22.1", default-features = false, features = ["webpki-tokio"], optional = true }
+bytes = { version = "1.0", optional = true }
+reqwest = { version = "0.11", optional = true, default-features = false }
 headers = { version = "0.3", optional = true }
 ## WS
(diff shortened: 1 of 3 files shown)


src/transports/http.rs
```diff
@@ -18,105 +22,59 @@ use std::{
...
-    url: hyper::Uri,
+    url: reqwest::Url,
     basic_auth: Option<HeaderValue>,
-    client: Client,
+    client: reqwest::Client,
 }
 
 impl Http {
     /// Create new HTTP transport connecting to given URL.
+    #[allow(unused_mut)]
```
tomusdrw, line 51: Couldn't this be moved right above of `let mut client_builder` line?

tomusdrw (commented):
That's awesome, thanks a lot! Could you please bump the version as well?

Cargo.toml
```diff
@@ -66,11 +64,13 @@ tokio-stream = { version = "0.1", features = ["net"] }
...
-http = ["_http_base", "hyper-proxy/tls"]
-http-tls = ["hyper-tls", "http"]
-http-rustls = ["hyper-rustls", "_http_base", "hyper-proxy/rustls-webpki"]
+wasm = ["js-sys", "wasm-bindgen", "wasm-bindgen-futures", "futures-timer/wasm-bindgen", "rand", "getrandom"]
+eip-1193 = ["wasm"]
+_http_base = ["reqwest", "bytes", "url", "base64", "headers"]
+http = ["_http_base"]
+http-tls = ["http", "reqwest/default-tls"]
+http-native-tls = ["http", "reqwest/native-tls"]
+http-rustls-tls = ["http", "reqwest/rustls-tls"]
```
tomusdrw, line 73: Would you mind adding two additional entries in `.github/workflows/rust.yml` to test these features work in isolation?

tomusdrw (commented):
Thanks again, I've just realised that it would be good to test newly introduced features on CI as well.

Cargo.toml
clouds56: I'm not familiar with Github action, tried and plz review.

clouds56:
Shall we run ci first?

tomusdrw (approved):

PR #497, opened 2021-06-04
Redo http transport

e00E:
The existing implementation has accumulated a lot of crust from initial
async support and use of hyper. Now that is using reqwest we can do many
things simpler.

Removed things that were needed for hyper but are handled already by
reqwest:
- setting the proxy
- setting the auth header
- setting the content type header
- setting content length header

Other changes:
- Move from struct with Future impl to BoxFuture. We were already using
BoxFuture internally anyawy.
- Log request and response as debug strings. This makes sure no control
characters like newlines end up in the output.
- Use from_utf8_lossy for the response log so that we can still
partially log non utf8 responses.
- Fix not handling batch responses according to the jsonrpc
specification which does not guarantee the ordering of the list.
- Add more context to error logs.

### DIFF
--- a/src/transports/http.rs
+++ b/src/transports/http.rs
@@ -8,7 +8,6 @@ use crate::{
 use futures::future::BoxFuture;
 #[cfg(feature = "wasm")]
 use futures::future::LocalBoxFuture as BoxFuture;
-use headers::HeaderValue;
 use jsonrpc_core::types::{Call, Output, Request, Value};
 use reqwest::{Client, Url};
 use serde::de::DeserializeOwned;
@@ -45,7 +44,7 @@ impl Http {
         let mut builder = Client::builder();
         #[cfg(not(feature = "wasm"))]
         {
-            builder = builder.user_agent(HeaderValue::from_static("web3.rs"));
+            builder = builder.user_agent(headers::HeaderValue::from_static("web3.rs"));
         }
         let client = builder
             .build()
--- a/src/transports/http.rs
+++ b/src/transports/http.rs
@@ -1,247 +1,172 @@
 //! HTTP Transport
 
-use crate::{error, helpers, rpc, BatchTransport, Error, RequestId, Transport};
+use crate::{
+    error::{Error, Result},
+    helpers, BatchTransport, RequestId, Transport,
+};
 #[cfg(not(feature = "wasm"))]
 use futures::future::BoxFuture;
 #[cfg(feature = "wasm")]
 use futures::future::LocalBoxFuture as BoxFuture;
-use futures::{
-    self,
-    task::{Context, Poll},
-    Future, FutureExt,
-};
-use reqwest::header::HeaderValue;
+use headers::HeaderValue;
+use jsonrpc_core::types::{Call, Output, Request, Value};
+use reqwest::{Client, Url};
+use serde::de::DeserializeOwned;
 use std::{
-    fmt,
-    ops::Deref,
-    pin::Pin,
+    collections::HashMap,
     sync::{
-        atomic::{self, AtomicUsize},
+        atomic::{AtomicUsize, Ordering},
         Arc,
     },
 };
-use url::Url;
 
-impl From<reqwest::Error> for Error {
-    fn from(err: reqwest::Error) -> Self {
-        Error::Transport(format!("{:?}", err))
-    }
-}
-
-impl From<reqwest::header::InvalidHeaderValue> for Error {
-    fn from(err: reqwest::header::InvalidHeaderValue) -> Self {
-        Error::Transport(format!("{}", err))
-    }
-}
-
-// The max string length of a request without transfer-encoding: chunked.
-const MAX_SINGLE_CHUNK: usize = 256;
-
-/// HTTP Transport (synchronous)
-#[derive(Debug, Clone)]
+/// HTTP Transport
+#[derive(Clone, Debug)]
 pub struct Http {
-    id: Arc<AtomicUsize>,
-    url: reqwest::Url,
-    basic_auth: Option<HeaderValue>,
-    client: reqwest::Client,
+    // Client is already an Arc so doesn't need to be part of inner.
+    client: Client,
+    inner: Arc<Inner>,
+}
+
+#[derive(Debug)]
+struct Inner {
+    url: Url,
+    id: AtomicUsize,
 }
 
 impl Http {
     /// Create new HTTP transport connecting to given URL.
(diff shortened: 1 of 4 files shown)


src/transports/http.rs
```diff
@@ -1,283 +1,176 @@
...
-        assert_eq!(http.basic_auth, Some(HeaderValue::from_static("Basic OnBhc3N3b3Jk")))
-    }
-
     async fn server(req: hyper::Request<hyper::Body>) -> hyper::Result<hyper::Response<hyper::Body>> {
         use hyper::body::HttpBody;
 
-        let expected = r#"{"jsonrpc":"2.0","method":"eth_getAccounts","params":[],"id":1}"#;
-        let response = r#"{"jsonrpc":"2.0","id":1,"result":"x"}"#;
+        let expected = r#"{"jsonrpc":"2.0","method":"eth_getAccounts","params":[],"id":0}"#;
+        let response = r#"{"jsonrpc":"2.0","id":0,"result":"x"}"#;
```
e00E, line 173: There is an old commit from several years ago that changed the initial id value from 0 to 1 in Http::new. I don't see why so I changed it back to 0.

src/transports/http.rs
```diff
@@ -1,283 +1,173 @@
...
+// Id is only used for logging.
+async fn execute_rpc<T: DeserializeOwned>(client: &Client, url: Url, request: &Request, id: RequestId) -> Result<T> {
+    log::debug!("[id:{}] sending request: {:?}", id, serde_json::to_string(&request)?);
+    let response = client
+        .post(url)
+        .json(request)
+        .send()
+        .await
+        .map_err(|err| Error::Transport(format!("failed to send request: {}", err)))?;
+    let status = response.status();
```
tomusdrw, line 74: What happened with:
- basic auth support
- custom user agent
- non-chunked request if they are below `MAX_SINGLE_CHUNK`
- proxy support
- tls support

Is it intended now for the user to use `with_client` method and pre-configure that?

src/transports/http.rs
```diff
@@ -1,283 +1,173 @@
...
+    }
+    helpers::arbitrary_precision_deserialize_workaround(&response)
+        .map_err(|err| Error::Transport(format!("failed to deserialize response: {}", err)))
+}
+
+type RpcResult = Result<Value>;
+
 impl Transport for Http {
-    type Out = Response<fn(Vec<u8>) -> error::Result<rpc::Value>>;
+    type Out = BoxFuture<'static, Result<Value>>;
```
tomusdrw, line 97: The goal of `type Out` on `Transport` was to avoid boxing, is it still possible to avoid?

src/transports/http.rs
tomusdrw: Well, it's only 12 months old, not several years :) I don't think there was any concrete reason to change that apart from aesthetics.

tomusdrw (commented):
Nice!

src/transports/http.rs
e00E: The previous code was not avoiding boxing. Probably an earlier version with hyper was but not currently. While `Response` did not have to be boxed, its field `state: ResponseState` was an enum of two boxes. So this code reduces the number of boxes by 1.
It probably would be possible to avoid boxing with a custom Future implementation again but considering we are sending network data, serializing data, storing a response in a Vector etc I do not think it is worth the complexity. The nice way to avoid boxing here in the future is with `type Out = impl Future<Item=Result<Value>>` but impl trait on associated types is unstable.
Anyway I would not like to do that as part of this PR as the code was already boxing but I will consider it for another one.

src/transports/http.rs
e00E: From glancing over the reqwest code it already automatically does basic auth and enables proxy support with the same env variable we were using. This is why I removed that part. Although I did not test this.

The ssl function the previous code was using does *not* enable ssl. It forces reqwest to use a specific ssl backend in case several of the ssl features are enabled. I don't see a reason to do this. If a user does enable multiple features it means they probably do not care and if they do care they can still customize the client.

> non-chunked request if they are below MAX_SINGLE_CHUNK

This one I didn't understand the purpose of and the code didn't give a reason. I believe it is better to let the http library decide how to send the request. It knows how long the data is so it can set Content Length as it pleases.

> custom user agent

I thought reqwest already set a user agent but now I see it does not. Will add the rust-web3 user agent back.

tomusdrw: Thanks for detailed explanations, I'm good with all of them.

The non-chunked request thing was mainly convenience - hyper was defaulting to send a chunked request (without `Content-Length`) and it was a bit more cumbersome to read over a raw TCP stream (chunk decoding). It's fine to let `reqwest` decide, I agree.

> I thought reqwest already set a user agent but now I see it does not. Will add the rust-web3 user agent back.

Cool, thanks!

src/transports/http.rs
tomusdrw: Fair enough, sounds good, the extra PR is optional if we were already `Box`ing.

tomusdrw (approved):
Looks good, thanks a lot! The WASM build requires a tiny fix.

e00E:
Looks like you can't set a user agent on wasm, which makes sense as the browser probably uses its own. Let's see if this fixes the build.

src/transports/http.rs
```diff
@@ -1,283 +1,184 @@
...
 use futures::future::BoxFuture;
 #[cfg(feature = "wasm")]
 use futures::future::LocalBoxFuture as BoxFuture;
-use futures::{
-    self,
-    task::{Context, Poll},
-    Future, FutureExt,
-};
-use reqwest::header::HeaderValue;
+use headers::HeaderValue;
```
tomusdrw, line 11:
```suggestion
```

src/transports/http.rs
```diff
@@ -1,283 +1,184 @@
...
         {
-            let proxy_env = std::env::var("HTTPS_PROXY");
-            client_builder = match proxy_env {
-                Ok(proxy_scheme) => {
-                    let proxy = reqwest::Proxy::all(proxy_scheme.as_str())?;
-                    client_builder.proxy(proxy)
-                }
-                Err(_) => client_builder.no_proxy(),
-            };
+            builder = builder.user_agent(HeaderValue::from_static("web3.rs"));
```
tomusdrw, line 48:
```suggestion
            builder = builder.user_agent(headers::HeaderValue::from_static("web3.rs"));
```

tomusdrw (approved):
Awesome, one last bit is WASM-build warning, and we're good to go :)

PR #509, opened 2021-06-27
Add url query params to the WS client 'resource' parameter if present

chris-ricketts:
Previously only the URL path section was provided when creating the WebSocket Client: 
```
Client::new(socket, host, url.path());
```
Now if there are query parameters in the URL they are also included in the `resource` parameter of `Client::new`


tomusdrw (approved):
Thanks a lot!

PR #514, opened 2021-07-02
Fix/513

### LINKED ISSUE #513: Receipt missing from/to fields
Not sure whether it should be included or not
Happy to make PR if it should
https://openethereum.github.io/JSONRPC-eth-module#eth_gettransactionreceipt

holygits:
closes #513


src/types/transaction.rs
```diff
@@ -66,6 +66,10 @@ pub struct Receipt {
     /// Number of the block this transaction was included within.
     #[serde(rename = "blockNumber")]
     pub block_number: Option<U64>,
+    /// Sender
+    pub from: Address,
```
tomusdrw, line 70: I wonder if we should use `serde-default` here to maintain backward compatibility with clients without `from` field. Since it wasn't here for a while, I suspect that not all clients might have been returning the data
```suggestion
    #[serde(default)]
    pub from: Address,
```

tomusdrw (approved):

tomusdrw (approved):
Thanks a lot!

Issue #523, opened 2021-07-16
Cannot run example: receipt can't be null after wait for confirmations


stnbu:
I have ganache-cli running with the correct seed words per the README file. When I run the example it complains:

```
$ cargo run --example contract
    Finished dev [unoptimized + debuginfo] target(s) in 0.12s
     Running `target/debug/examples/contract`
thread 'main' panicked at 'receipt can't be null after wait for confirmations; qed', /Users/mburr/git/rust-web3/src/confirm.rs:88:10
stack backtrace:
   0: rust_begin_unwind
             at /rustc/9bc8c42bb2f19e745a63f3445f1ac248fb015e53/library/std/src/panicking.rs:493:5
   1: core::panicking::panic_fmt
             at /rustc/9bc8c42bb2f19e745a63f3445f1ac248fb015e53/library/core/src/panicking.rs:92:14
   2: core::option::expect_failed
             at /rustc/9bc8c42bb2f19e745a63f3445f1ac248fb015e53/library/core/src/option.rs:1321:5
   3: core::option::Option<T>::expect
             at /rustc/9bc8c42bb2f19e745a63f3445f1ac248fb015e53/library/core/src/option.rs:349:21
   4: web3::confirm::send_transaction_with_confirmation_::{{closure}}
             at ./src/confirm.rs:85:19
   5: <core::future::from_generator::GenFuture<T> as core::future::future::Future>::poll
             at /rustc/9bc8c42bb2f19e745a63f3445f1ac248fb015e53/library/core/src/future/mod.rs:80:19
   6: web3::confirm::send_transaction_with_confirmation::{{closure}}
             at ./src/confirm.rs:103:5
   7: <core::future::from_generator::GenFuture<T> as core::future::future::Future>::poll
             at /rustc/9bc8c42bb2f19e745a63f3445f1ac248fb015e53/library/core/src/future/mod.rs:80:19
   8: web3::contract::deploy::Builder<T>::do_execute::{{closure}}
             at ./src/contract/deploy.rs:212:23
   9: <core::future::from_generator::GenFuture<T> as core::future::future::Future>::poll
             at /rustc/9bc8c42bb2f19e745a63f3445f1ac248fb015e53/library/core/src/future/mod.rs:80:19
  10: web3::contract::deploy::Builder<T>::execute::{{closure}}
             at ./src/contract/deploy.rs:58:9
  11: <core::future::from_generator::GenFuture<T> as core::future::future::Future>::poll
[...]

Issue #432, opened 2020-12-21
Possible regression from 0.13 to 0.14


elpiel:
We have a test that started failing after upgrading to `0.14` that was working before.

The test and the failing call can be found here:
https://github.com/AdExNetwork/adex-validator-stack-rust/blob/dev/adapter/src/ethereum.rs#L591-L600

The error I get is:
```
---- ethereum::test::should_validate_valid_channel_properly stdout ----
thread 'ethereum::test::should_validate_valid_channel_properly' panicked at
'Correct parameters are passed to the constructor.: Abi(Other("hex decode error: Odd number of digits"))',
adapter/src/ethereum.rs:613:14
```

Which after a quick check in the repo is found here:

0.13:
https://github.com/tomusdrw/rust-web3/blob/v0.13.0/src/contract/deploy.rs#L131

0.14:
https://github.com/tomusdrw/rust-web3/blob/v0.14.0/src/contract/deploy.rs#L126

elpiel:
I believe this is caused by a new line `\n` at the end of the json file we read and use in our deploy. This was working on 0.13

PR #524, opened 2021-07-16
`trim_right()` the result of `include_str!("./res/contract_token.code")`

stnbu:
I have encountered #432 as described in #523

Since the presence of a `\n` can vary depending on tooling, fixing it in the example code.


examples/contract.rs
```diff
@@ -12,7 +12,10 @@ async fn main() -> web3::contract::Result<()> {
 
     let my_account = hex!("d028d24f16a8893bd078259d413372ac01580769").into();
     // Get the contract bytecode for instance from Solidity compiler
-    let bytecode = include_str!("./res/contract_token.code");
+    let bytecode = include_str!("./res/contract_token.code")
+        .to_string()
+        .trim_end()
+        .to_owned();
```
tomusdrw, line 18:
```suggestion
    let bytecode = include_str!("./res/contract_token.code").trim_end();
```
Is it really necessary to convert to an owned `String`? I think `&str` should work just fine.

tomusdrw (approved):

PR #529, opened 2021-07-25
Adding in CallRequestBuilder and tests

### LINKED ISSUE #490: Add `CallRequest` and `TransactionRequest` builders
Currently it's a bit cumbersome to define these structs, you can always do:
```
CallRequest {
  gas: Some(...),
  .. Default::default(),
}
```
but it still might be more convenient to have a builder.

bytesoul:
Following up on the Enhancement mentioned in #490. If this is acceptable I will also add a TransactionRequestBuilder.
resolves #490

### DIFF
--- a/src/types/transaction_request.rs
+++ b/src/types/transaction_request.rs
@@ -143,7 +143,7 @@ pub struct TransactionRequest {
     pub access_list: Option<AccessList>,
 }
 
-impl  TransactionRequest {
+impl TransactionRequest {
     /// Funtion to return a builder for a Transaction Request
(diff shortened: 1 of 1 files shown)


tomusdrw (commented):
Looks cool! It would be awesome to add docs and properly format the code.

Also have you considered just using some proc-macro library to generate the builder? Maybe something like this: https://github.com/colin-kiegel/rust-derive-builder

bytesoul:
Let me know if the formatting is correct now

tomusdrw:
@bytesoul please simply use `rustfmt` and run `cargo fmt`.

PR #533, opened 2021-08-11
EIP-1559 Support

mdben1247:


tomusdrw (commented):
Great, thanks! Could you please address the build issues?

mdben1247:
I see `base_fee_per_gas` field has been added after I forked, so there are now two instances of this field in this merge. Any hints how to resolve that?

mdben1247:
So, this evolved a bit. There is added support for signing EIP-1559 transactions directly from rust-web3 and field `gas_price` in `Transaction` is now optional.

src/api/accounts.rs
```diff
@@ -82,21 +87,37 @@ mod accounts_signing {
                 };
             }
             let from = key.address();
+
+            let gas_price = match tx.transaction_type {
+                Some(tx_type) if tx_type == U64::from(2) && tx.max_fee_per_gas.is_some() => tx.max_fee_per_gas,
```
tomusdrw, line 92: Can we move the `U64::from(2)` or just `2` into a constant?

src/api/accounts.rs
```diff
@@ -165,74 +186,172 @@ mod accounts_signing {
...
+                stream.begin_list(access.storage_keys.len());
+                for storage_key in access.storage_keys.iter() {
+                    stream.append(storage_key);
+                }
+            }
+        }
+
+        fn encode(&self, chain_id: u64, signature: Option<&Signature>) -> Vec<u8> {
+            match self.transaction_type.map(|t| t.as_u64()) {
+                // type 0: Legacy
```
tomusdrw, line 302: I'd prefer these types to be either an `enum` (converted into `u8/u16/u64`) or a set of constants.

src/signing.rs
```diff
@@ -113,6 +117,19 @@ mod feature_gated {
             Ok(Signature { v, r, s })
         }
 
+        fn sign_message(&self, message: &[u8]) -> Result<Signature, SigningError> {
+            let message = Message::from_slice(&message).map_err(|_| SigningError::InvalidMessage)?;
+            let (recovery_id, signature) = Secp256k1::signing_only()
```
tomusdrw, line 122: Can you use the newly introduced shared `CONTEXT` here?

mdben1247: Uses the new context now.

src/api/accounts.rs
mdben1247: Added constants.

src/api/accounts.rs
mdben1247: Constant.

tomusdrw (approved):
Awesome, thanks!

Issue #534, opened 2021-08-13
EC Recover is slow


Recmo:
The [`recover`](https://docs.rs/web3/0.17.0/web3/signing/fn.recover.html) function is unnecessarily slow. On my machine it takes about 7 ms instead of 60 μs.

The cause is [this line](https://docs.rs/web3/0.17.0/src/web3/signing.rs.html#128):

```rust
        let public_key = Secp256k1::verification_only()
            .recover(&message, &signature)
```

This calls `Secp256k1::verification_only()` for each recover, creating a new context and recomputing all the lookup tables each time. This is documented in [`secp256k1`](https://docs.rs/secp256k1/0.20.3/secp256k1/):

> [...] the library uses context objects that contain precomputation tables which are created on object construction. Since this is a slow operation (10+ milliseconds, vs ~50 microseconds for typical crypto operations, on a 2.70 Ghz i7-6820HQ) the tables are optional, giving a performance boost for users who only care about signing, only care about verification, or only care about parsing [...]

So we should statically allocate the context and initialize it once:

```rust
use once_cell::sync::Lazy;
use secp256k1::{
    recovery::{RecoverableSignature, RecoveryId},
    Error, Message, Secp256k1, VerifyOnly,
};
use sha3::{Digest, Keccak256};
use web3::types::Address;

static CONTEXT: Lazy<Secp256k1<VerifyOnly>> = Lazy::new(Secp256k1::verification_only);

pub fn recover(message: &[u8], signature: &[u8], recovery_id: i32) -> Result<Address, Error> {
    // Recover public key
    let message = Message::from_slice(message)?;
    let recovery_id = RecoveryId::from_i32(recovery_id)?;
    let signature = RecoverableSignature::from_compact(signature, recovery_id)?;
    let public_key = CONTEXT.recover(&message, &signature)?;

    // Hash public key into address
    let public_key = public_key.serialize_uncompressed();
    debug_assert_eq!(public_key[0], 0x04);
    let hash = {
        let mut hasher = Keccak256::new();
        hasher.update(&public_key[1..]);
        hasher.finalize()
    };
    let address = Address::from_slice(&hash[12..]);
    Ok(address)
}
```

Benchmarking the above shows a 100x improvement:

```
web3_recover            time:   [7.0468 ms 7.1243 ms 7.2109 ms]                         
my_recover              time:   [58.773 us 63.094 us 67.377 us]                          
[...]

PR #535, opened 2021-08-14
Initialize secp256k1 context only once

tomusdrw:
As suggested in #534

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -23,12 +23,14 @@ jsonrpc-core = "18.0.0"
 log = "0.4.6"
 parking_lot = "0.11.0"
 rlp = "0.5"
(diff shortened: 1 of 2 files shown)


Recmo (commented):
Note that the Secp256k1 context is only intialized once for verifying. For signing (and pubkey derivation) it still intializes contexts each time:

https://github.com/tomusdrw/rust-web3/pull/535/files#diff-e30da8c800b745948f33e34e23570b0a94f7a6f90a0b26433e851dc99f27267cR100

https://github.com/tomusdrw/rust-web3/pull/535/files#diff-e30da8c800b745948f33e34e23570b0a94f7a6f90a0b26433e851dc99f27267cLR153

PR #548, opened 2021-09-24
Optional Provider

### LINKED ISSUE #545: EIP-1193 crash if no ethereum object
Would be cool if it returned a nice error message when the user has no compatible wallet installed with their browser instead of crashing.

SionoiS:
Fix #545

Makes the EIP-1193 provider optional forcing handling of the case where there's no compatible wallet.


tomusdrw (approved):

tomusdrw:
Thanks a lot!

PR #549, opened 2021-09-28
Fix panic when MetaMask returns error

### LINKED ISSUE #544: eip_1193: entered unreachable code when canceling transaction in MetaMask
```
panicked at 'internal error: entered unreachable code: Unable to parse request response: Err(Err(Error("unknown field `stack`, expected one of `code`, `message`, `data`", line: 1, column: 90)))'
```
Code to reproduce this behavior:
```rust
let window = web_sys::window().expect("no window available");
let ethereum = window.get("ethereum").expect("MetaMask is not installed");
let eth = ethereum.unchecked_into::<web3::transports::eip_1193::Provider>();
let web3 = web3::Web3::new(web3::transports::eip_1193::Eip1193::new(eth)));
web3.eth().request_accounts().await;
```
Run and click "Cancel" when MetaMask will offer to connect your wallet to website.
The same bug appears when cancelling transaction confirmation.

dvec:
Fix #544

### DIFF
--- a/src/transports/eip_1193.rs
+++ b/src/transports/eip_1193.rs
@@ -13,7 +13,7 @@ use crate::{
 use futures::{channel::mpsc, future::LocalBoxFuture, Stream};
 use jsonrpc_core::{
     error::{Error as RPCError, ErrorCode as RPCErrorCode},
-    types::request::{Call, MethodCall}
+    types::request::{Call, MethodCall},
 };
 use serde::{
     de::{value::StringDeserializer, IntoDeserializer},
@@ -249,7 +249,7 @@ impl Provider {
             /// Optional data
             pub data: Option<serde_json::value::Value>,
             /// Optional stack
-            pub stack: Option<serde_json::value::Value>
+            pub stack: Option<serde_json::value::Value>,
         }
 
         impl Into<RPCError> for RPCErrorExtra {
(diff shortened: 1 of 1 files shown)


tomusdrw (commented):
Nice, thanks! Any chance we can cover this behaviour with a test case?

dvec:
We can't run tests directly in browser, so I think the only solution is to mock `window.ethereum`, which may be a tricky task

tomusdrw (commented):
What if we extract response parsing to a standalone (rust) function and just unit test that? I.e. that given the expected payload it parses correctly. wdyt?

dvec:
Sounds like a solution, I'll try to commit it today

dvec:
I'm wondering why there is a panic at all in case of invalid response while we have `web3::error::Error::InvalidResponse`. Especially considering that `std::panic::catch_unwind` is not available on `wasm` target (what also means that it's impossible to use `#[should_panic]` in tests). Can we refuse this behavior and return `Err(web3::error::Error::InvalidResponse(...))` instead of panic?

tomusdrw:
>  Can we refuse this behavior and return Err(web3::error::Error::InvalidResponse(...)) instead of panic?

Yeah, I just looked at the code and indeed we shouldn't be panicking and definitely the code is not `unreachable` as we can tell. If you want to bundle the fix in the same PR I'm fine with that.

tomusdrw (approved):
Thanks a lot!

PR #572, opened 2021-11-23
eth_getProof balance and nonce to U256

lightyear15:
after some real-world testing, we figured balance as well as nonce
fields might reach the upper limit of U64.
Moving these fields to U256


lightyear15:
sorry about this :disappointed: 
should have then more tests before my first PR

tomusdrw:
No worries! Could you please add a testcase for these large values as well?

lightyear15:
Hi @tomusdrw 
I changed the unit test as you requested. 
The proof does not need to be consistent as that is not the purpose of the test. So I just changed the value of ``balance`` and ``nonce`` fields to make them more similar to a real-case.

tomusdrw:
Perfect, thank you!

tomusdrw (approved):

PR #581, opened 2021-12-29
signed_call (without confirmations)

dpuyosa:
Sometimes you just need the tx_hash and no confirmations. ``signed_call_with_confirmations`` forces you to wait for at least 1 confirmation (0 does panic #523).

Follows having ``my_contract.call()`` and ``my_contract.call_with_confirmations()``.

### DIFF
--- a/src/contract/mod.rs
+++ b/src/contract/mod.rs
@@ -383,7 +383,10 @@ mod contract_signing {
             self.eth.send_raw_transaction(signed.raw_transaction).await
         }
 
-        /// Execute a signed contract function and wait for confirmations
+        /// Submit contract call transaction to the transaction pool and wait for the transaction to be included in a block.
+        ///
+        /// This function will wait for block inclusion of the transaction before returning.
+        // If you'd rather just submit transaction and receive it's hash, please use [`signed_call`] instead.
         pub async fn signed_call_with_confirmations(
             &self,
             func: &str,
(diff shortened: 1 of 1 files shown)


src/contract/mod.rs
```diff
@@ -365,7 +365,33 @@ mod contract_signing {
             if let Some(value) = options.value {
                 tx.value = value;
             }
-            let signed = accounts.sign_transaction(tx, key).await?;
+            accounts.sign_transaction(tx, key).await
+        }
+
+        /// Execute a signed contract function
```
tomusdrw, line 371:
```suggestion
        /// Submit contract call transaction to the transaction pool.
        ///
        /// Note this function DOES NOT wait for any confirmations, so there is no guarantees that the call is actually executed.
        /// If you'd rather wait for block inclusion, please use [`signed_call_with_confirmations`] instead.
```

src/contract/mod.rs
```diff
@@ -365,7 +365,33 @@ mod contract_signing {
...
+            func: &str,
+            params: impl Tokenize,
+            options: Options,
+            key: impl signing::Key,
+        ) -> crate::Result<H256> {
+            let signed = self.sign(func, params, options, key).await?;
+            self.eth.send_raw_transaction(signed.raw_transaction).await
+        }
+
+        /// Execute a signed contract function and wait for confirmations
```
tomusdrw, line 383:
```suggestion
        /// Submit contract call transaction to the transaction pool and wait for the transaction to be included in a block.
        ///
        /// This function will wait for block inclusion of the transaction before returning.
        // If you'd rather just submit transaction and receive it's hash, please use [`signed_call`] instead.
```

tomusdrw (approved):
lgtm, thanks!

PR #585, opened 2022-01-14
Report HTTP failure codes as RPC errors if returned for an RPC call

marmistrz:
At least in the Alchemy JSON RPC API over HTTP, rate limit errors are returned as HTTP 429. Currently these errors are converted to Error::Transport(String) and it's not possible to cleanly access the status code. Moreover, `Error::Transport` also means errors like failure to deserialize, whose handling will be completely different than handling of capacity limits or rate limit.

This PR changes representations of the errors returned in `execute_rpc` to `rpc::error::Error`. Since the HTTP error codes aren't any of the variants defined by jsonrpc-core, the error code will be `ErrorCode::ServerError(i64)`.

cf. https://docs.alchemy.com/alchemy/documentation/error-reference

### DIFF
--- a/src/transports/http.rs
+++ b/src/transports/http.rs
@@ -1,19 +1,14 @@
 //! HTTP Transport
 
 use crate::{
-    error::{Error, Result},
-    helpers,
-    rpc::error::Error as RPCError,
-    BatchTransport, RequestId, Transport,
+    error::{Error, Result, TransportError},
+    helpers, BatchTransport, RequestId, Transport,
 };
 #[cfg(not(feature = "wasm"))]
 use futures::future::BoxFuture;
 #[cfg(feature = "wasm")]
 use futures::future::LocalBoxFuture as BoxFuture;
-use jsonrpc_core::{
-    types::{Call, Output, Request, Value},
-    ErrorCode,
-};
+use jsonrpc_core::types::{Call, Output, Request, Value};
 use reqwest::{Client, Url};
 use serde::de::DeserializeOwned;
 use std::{
@@ -53,7 +48,7 @@ impl Http {
         }
         let client = builder
             .build()
-            .map_err(|err| Error::Transport(format!("failed to build client: {}", err)))?;
+            .map_err(|err| Error::Transport(TransportError::Message(format!("failed to build client: {}", err))))?;
         Ok(Self::with_client(client, url.parse()?))
     }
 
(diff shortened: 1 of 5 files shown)


src/transports/http.rs
```diff
@@ -86,16 +91,15 @@ async fn execute_rpc<T: DeserializeOwned>(client: &Client, url: Url, request: &R
         .bytes()
         .await
         .map_err(|err| Error::Transport(format!("failed to read response bytes: {}", err)))?;
-    log::debug!(
-        "[id:{}] received response: {:?}",
-        id,
-        String::from_utf8_lossy(&response).as_ref()
-    );
+    let decoded_response = String::from_utf8_lossy(&response);
```
tomusdrw, line 94: I'd prefer if this decoding was only done if:
1. we print debug messages,
2. or we are in the error scenario.

Currently it will be converted to a string in a "happy path" as well, but the result is dropped.
I realize that in case of both (1) & (2) conditions being true, we might end up decoding twice, but my guts tell me this would be a better trade-off (happy to be proven otherwise with a benchmark though).

src/transports/http.rs
```diff
@@ -86,16 +91,15 @@ async fn execute_rpc<T: DeserializeOwned>(client: &Client, url: Url, request: &R
...
+    log::debug!("[id:{}] received response: {:?}", id, decoded_response.as_ref());
     if !status.is_success() {
-        return Err(Error::Transport(format!(
-            "response status code is not success: {}",
-            status
-        )));
+        let code = ErrorCode::from(i64::from(status.as_u16()));
+        return Err(Error::Rpc(RPCError {
+            code,
```
tomusdrw, line 99: Not a fan of mixing HTTP error codes (i.e. transport specific) with the ones returned by actual JSON-RPC server. What if we have a conflict and conflate an JSON-RPC error code with some HTTP error code?

I'd rather introduce a richer `Transport` variant, with following fields:
```rust
struct TransportError {
  /// Transport implementation self-identification.
  /// 
  /// This can be used to determine the actual transport (HTTP, WS, etc). being used if it's not statically known.
  transport_id: String,
  /// Transport-specific error code.
  code: u64,
  /// Arbitrary, developer-readable description of the occurred error.
  message: String,
}
```
wdyt?

tomusdrw (changes requested):

src/transports/http.rs
marmistrz: JSON-RPC codes are to be negative, http/ws error codes are positive so there shouldn't be a clash, so I'm fine with both approaches. It makes some sense to use a different variant for transport and rpc errors because the former are much more retriable than the latter in many cases.

This was roughly my initial idea, however, `Transport` is not used in a compatible way right now as it's also used to report errors like https://github.com/tomusdrw/rust-web3/blob/033ac37fff3990a943b9362608cd2c151914f3e9/src/transports/http.rs#L51 or https://github.com/tomusdrw/rust-web3/blob/033ac37fff3990a943b9362608cd2c151914f3e9/src/transports/http.rs#L88. 

We could either introduce a new variant or use `code: Option<u64>`.

src/transports/http.rs
marmistrz: I agree it's a better tradeoff, good point!

src/transports/http.rs
tomusdrw: > JSON-RPC codes are to be negative, http/ws error codes are positive so there shouldn't be a clash

JSON-RPC codes can be anything actually, the spec just defines a few reserved ranges which are negative.

> This was roughly my initial idea, however, Transport is not used in a compatible way right now as it's also used to report errors like

We can either assign a `code` to each of such error (and expose constants with that values), use a "catch-all" `0` code for that or use `Option<u64>` as you suggested - I'm fine with either.

marmistrz: I did a slightly different implementation given that we're unlikely to get a developer-defined message and a code. I also skipped the `transport_id` because it would be a lot of boilerplate.

marmistrz:
Ping, what do you think about the new implementation?

src/transports/mod.rs
```diff
@@ -1,6 +1,8 @@
 //! Supported Ethereum JSON-RPC transports.
 
 pub mod batch;
+use crate::error::TransportError;
```
tomusdrw, line 4: Can you feature gate the import or use a direct path to prevent warnings?

src/transports/ws.rs
```diff
@@ -143,16 +156,10 @@ impl WsServerTask {
         let (sender, receiver) = match handshake.await? {
             ServerResponse::Accepted { .. } => client.into_builder().finish(),
             ServerResponse::Redirect { status_code, location } => {
```
tomusdrw, line 158:
```suggestion
            ServerResponse::Redirect { status_code, ... } => {
```
to prevent warnings

tomusdrw (approved):
Looks good! Please fix the tiny warnings that can be seen during the CI and we're good to go.

Issue #567, opened 2021-11-08
websocket infura secret support


Mirko-von-Leipzig:
I'm not too clued up on websockets and authentication. Is it currently possible to utilise websockets along with password / username? 

Creating a websocket currently takes in just `url: &str` which doesn't allow setting a username and password.

tomusdrw:
Would `wss://username:password@host/` work? If not we can easily add authorisation header support (AFAICT Infura is using basic auth).

Mirko-von-Leipzig:
`wss://username:password@host/` gets rejected -- unless I'm doing it wrong: `wss://:<secret>@host` (since there is no username)? This form gets rejected.

The example given in the docs is:
`wscat -c wss://mainnet.infura.io/ws/v3/YOUR-PROJECT-ID --auth ":YOUR-PROJECT-SECRET"`

PR #590, opened 2022-01-26
Allow setting basic auth credentials in websocket url

717a56e1:
These could be set before, but would be ignored by soketto. Now they're propagated to the backend.

### DIFF
--- a/src/transports/ws.rs
+++ b/src/transports/ws.rs
@@ -139,6 +139,20 @@ impl WsServerTask {
             resource
         );
         let mut client = Client::new(socket, host, &resource);
+        let maybe_encoded = url.password().map(|password| {
+            use headers::authorization::{Authorization, Credentials};
+            log::trace!("Basic auth username: {}", url.username());
+            log::trace!("Basic auth password: {}", password);
+            Authorization::basic(url.username(), password).0.encode().to_str().unwrap().as_bytes().to_owned()
+        });
+
+        let headers = if let Some(ref head) = maybe_encoded {
+            Some([soketto::handshake::client::Header {name: "Authorization",value: &head }])
+        } else { None };
+
+        if let Some(ref head) = headers {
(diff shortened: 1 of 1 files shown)


src/transports/ws.rs
```diff
@@ -139,6 +139,20 @@ impl WsServerTask {
             resource
         );
         let mut client = Client::new(socket, host, &resource);
+        let maybe_encoded = url.password().map(|password| {
+            use headers::authorization::{Authorization, Credentials};
+            log::trace!("Basic auth username: {}", url.username());
+            log::trace!("Basic auth password: {}", password);
```
tomusdrw, line 145:
```suggestion
```

Is this really useful?

src/transports/ws.rs
```diff
@@ -139,6 +139,20 @@ impl WsServerTask {
...
+            Authorization::basic(url.username(), password).0.encode().to_str().unwrap().as_bytes().to_owned()
```
tomusdrw, line 146:
```suggestion
            Authorization::basic(url.username(), password).0.encode().as_bytes().to_vec()
```
I don't think it's necessary to go via `&str` first, especially given we just unwrap.

src/transports/ws.rs
```diff
@@ -139,6 +139,20 @@ impl WsServerTask {
...
+        });
+
+        let headers = if let Some(ref head) = maybe_encoded {
+            Some([soketto::handshake::client::Header {name: "Authorization",value: &head }])
```
tomusdrw, line 150:
```suggestion
            Some([soketto::handshake::client::Header {name: "Authorization", value: &head }])
```
Please make sure to format the code with rustfmt.

tomusdrw (commented):
Great catch, thanks a lot!

tomusdrw:
Fixes https://github.com/tomusdrw/rust-web3/issues/567

Issue #593, opened 2022-02-04
Cannot send EIP-1559 Contract Interactions


the-Jamz:
## Problem

When setting transaction type to 2, which should according to PR #533 use EIP-1559, I'm still sending a legacy contract interaction. I was however able to send EIP-1559 transactions just fine.

## Perhaps a misunderstanding

Is this the fault of the contract? I'm new to this in general, so maybe I misunderstand, but I would expect that the client is responsible for sending the contract interaction and therefore it shouldn't be my test smart contract causing the issue? If I'm mistaken, I hope someone else can at least learn from this issue at some point.

## Possible duplicates

I did find issue #542 but the solution is not clear.

## Setup

~~I'm using revision [e81265bb4a3f40a5ac2d60bd2f57f0fce0b5af76](https://github.com/tomusdrw/rust-web3/commit/e81265bb4a3f40a5ac2d60bd2f57f0fce0b5af76) because PR #591 seems to have created a conflict with tokio.~~
`cargo update` to the rescue

To be precise: `web3 = { git = "https://github.com/tomusdrw/rust-web3", features = ["default"] }`

I have tried playing around with feature flags, maybe I missed something, but couldn't get it to work.

I'm testing using an NFT project on the rinkeby testnet created with [this guide](https://ethereum.org/en/developers/tutorials/how-to-write-and-deploy-an-nft/). You can view mine [here](https://rinkeby.etherscan.io/address/0x1038d3e4ee2dee4d03d2e079c9b57827c2593d21).

## Code

```Rust
    let options = web3::contract::Options {
        transaction_type: Some(2.into()),
        ..Default::default()
    };

    let tx = contract
        .signed_call_with_confirmations(
            "mintNFT",
            (our_address, "".to_string()),
            options,
            1,
            &prvk
        )
        .await
        .unwrap();
```

717a56e1:
#600 should fix this.

the-Jamz:
PR #600 was merged resolving the issue.

PR #600, opened 2022-02-07
Add omitted optional fields to signed call

717a56e1:


the-Jamz:
Can confirm this fixes my problem in issue #593! Thanks for figuring this out.

tomusdrw (approved):

tomusdrw:
Thanks!

PR #609, opened 2022-02-22
Update ethereum-types dependency to version 0.13.1

jeffreybolle:


jeffreybolle:
Sorry, this does not make sense until ethabi picks up ethereum-types version 0.13.1 first.

PR #616, opened 2022-03-07
Make fields optional when deserializing for compatibility

leoyvens:
This changes are necessary for compatibility with some EVM-compatible chains such as Celo and Fantom. I hope you're willing to take this small patch to support these chains.


tomusdrw:
@leoyvens Thanks! Do you mind feature-gating this and adding a feature description in the README? I'd prefer to keep the library strict by default.

leoyvens:
@tomusdrw Thank you for the review, I've added a feature gate.

tomusdrw (approved):

PR #626, opened 2022-04-12
Make debug output for Bytes hex strings (0xdeadbeef)

lutter:
This makes the debug representation of `Bytes` much more readable than the default which represents it as a `Vec<u8>`


tomusdrw (approved):
Awesome, thanks!

PR #630, opened 2022-05-02
Make Transaction from account signing public

Deniallugo:
# Problem
It's impossible to use web3 library for signing transactions without using any transport.

# Description  
In our project, we use the signer regardless of the transport and fill all necessary fields by our logic. We do not need to go to the transport to get nonce and gas price. But we want to use `Transaction` for signing because rlp encoding/decoding is not trivial and it's easy to make a mistake here.

# Solution 
 I made the necessary struct public it allows us to use just a private key for the signing structure, where all necessary fields are filled.

### DIFF
--- a/src/api/accounts.rs
+++ b/src/api/accounts.rs
@@ -36,7 +36,7 @@ impl<T: Transport> Accounts<T> {
 }
 
 #[cfg(feature = "signing")]
-mod accounts_signing {
+pub(crate) mod accounts_signing {
     use super::*;
     use crate::{
         api::Web3,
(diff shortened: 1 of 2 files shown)


tomusdrw (changes requested):
Hi @Deniallugo and sorry for the delay!

While I'm okay with the idea of exposing the `Transaction` type, I don't like the location where it ends up being exposed and that it is in fact a re-export.

Would you mind moving the `Transaction` type and all it's implementations to the `signing` module, which is already public? (instead of exposing it from `api`).

PR #633, opened 2022-05-09
Re-export secp256k1::SecretKey.

dandanlen:
`secp256k1::SecretKey` is required in order to use the offline signing functionality but at present it's not re-exported. This means anyone who want to use the signing functionality has to add an explicit dependency on `secp256k1`, and has to keep it in-sync with the version in `web3`.

This PR changes the re-export of the `secp256k1::SecretKey` from `pub(crate)` to `pub` so that is can be accessed through the `web3::signing` module, removing the requirement for an explicit dependency on secp256k1.

### DIFF
--- a/examples/transaction_public.rs
+++ b/examples/transaction_public.rs
@@ -1,10 +1,9 @@
 use std::str::FromStr;
 
-use secp256k1::SecretKey;
-
 use web3::{
     ethabi::ethereum_types::U256,
     types::{Address, TransactionParameters},
+    signing::SecretKey,
 };
 
 /// Below generates and signs a transaction offline, before transmitting it to a public node (eg Infura)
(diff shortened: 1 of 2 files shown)


tomusdrw:
While I think the original intent was to only expose `trait Key` to prevent misusing `SecretKey` (i.e. using it without zeroize), I believe it isn't sufficient to prevent misuse and in reality handling the secret keys properly is way beyond the purpose of this crate.
The `secp256k1::SecretKey` documentation also mentions that keys should be handled and `SecretKey::non_secure_erase` method is exposed, linking to `zeroize` crate for more info, hence I'm willing to merge the PR to improve usability.

tomusdrw (approved):

Issue #566, opened 2021-11-05
Got response from unknown request


MrCasCode:
Hi, I'm trying to subscribe to events in a contract:

```
   let ws = web3::transports::WebSocket::new("mynode").await?;
   let web3 = web3::Web3::new(ws);
 
   let contract = Contract::from_json(
        web3.eth(),
        hex!("4e68Ccd3E89f51C3074ca5072bbAC773960dFa36").into(),
        include_bytes!("../contracts/uniswap_v3_pool.json"),
    )?;

   filter = FilterBuilder::default()
        .address(vec![contract.address()])
        .from_block(BlockNumber::Latest)
        .build();

    let mut evsub = web3ws.eth_subscribe().subscribe_logs(filter).await?;

    println!("Sub id {:?}", evsub.id());

    loop {
        let r = evsub.next().await;
        match r {
            Some(ev) => println!("{:?}", ev?),
            None => ()
        }
    }

```

And everytime an event happens (checking on etherscan) I receive a response from unknown request

```
[2021-11-05T08:14:40Z DEBUG web3::transports::ws] [1] Calling: {"jsonrpc":"2.0","method":"eth_subscribe","params":["logs",{"address":"0x4e68ccd3e89f51c3074ca5072bbac773960dfa36","fromBlock":"latest"}],"id":1}
Sub id SubscriptionId("0x10ff0e1299f046edb86e5c774741102fa889ab7e15dd")
[2021-11-05T08:14:58Z WARN  web3::transports::ws] Got response for unknown request (id: 0)
[2021-11-05T08:15:32Z WARN  web3::transports::ws] Got response for unknown request (id: 0)
```

The same is happening when I try to run the examples subscribing to new heads. 

Any idea of what could be happening?
Thanks.

MrCasCode:
If that helps I can see in `ws.rs:259`:

`response = Error("data did not match any variant of untagged enum Response", line: 0, column: 0)`

MrCasCode:
Still digging trying to understand why the websocket response fails to be decoded whereas the http works. 

Received http:
```json
{
    "jsonrpc": "2.0",
    "id": 93,
    "result": [
        {
            "address": "0x88e6a0c2ddd26feeb64f039a2c41296fcb3f5640",
            "blockHash": "0x88cd879b0d64a1a182148424e033498c2fef0903133dca66364fdf5fa41202b8",
            "blockNumber": "0xcef012",
[...]

PR #639, opened 2022-05-27
Report response parse errors to avoid freezes

marmistrz:
This is at least a workaround for the issue reported in #566. 

Whether or not we choose to be strict in parsing, we should report the deserialization errors instead of letting the called wait indefinitely.

### DIFF
--- a/src/transports/ws.rs
+++ b/src/transports/ws.rs
@@ -262,7 +262,7 @@ enum ParsedMessage {
     Notification(rpc::Notification),
     /// Represents a valid JSON-RPC response
     Response(rpc::Response),
-    /// Represents an invalid JSON-RPC response
+    /// Represents an malformed JSON-RPC response, although containing the request `id`
     InvalidResponse(rpc::Id),
 }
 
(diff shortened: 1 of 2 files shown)


src/transports/ws.rs
```diff
@@ -257,56 +257,94 @@ fn as_data_stream<T: Unpin + futures::AsyncRead + futures::AsyncWrite>(
...
+    if let Ok(notification) = helpers::to_notification_from_slice(data) {
+        Some(ParsedMessage::Notification(notification))
+    } else if let Ok(response) = helpers::to_response_from_slice(data) {
+        Some(ParsedMessage::Response(response))
+    } else if let Some(id) = helpers::response_id_from_slice(data) {
+        Some(ParsedMessage::InvalidResponse(id))
+    } else {
+        None
+    }
+}
```
tomusdrw, line 279: I guess it might be better to simply use untagged enum `ParsedMessage` with default serde deserialize behaviour? Given some tests are written I think it should be simpler to reason about and possibly also faster than manually attempting to parse the response couple of times.

tomusdrw (changes requested):

Issue #650, opened 2022-06-24
Getting txHash along with the contract logs 


r0wdy1:
As of now (0.18.0) there are some flaws with Contract.events method :

- there seems to be no way to get the hash of transaction associated wit a specific event ( although the original spec and the log struct model has a field and the data itself is actually retrieved from the node), which makes it impossible to get CallData for a specific emitted event instance
- blockFrom, blockTo, blockHash and address filled are hardcoded as blank, so from my experience such a query wouldn't be executed by Infura at all 
- Corresponding builder functions for  blockFrom, blockTo, blockHash fields receive values, not options, so there is additional complexity when creating filter

I tried my best ( please be patient, this is my first PR)  to address aforementioned issues, may be not in a best way. I'm not really excited about making a new type for events with meta, the current idea with tuples, that matches event signature, seems really elegant, but unfortunately the txHash is actually necessary is lot's of cases

PR #651, opened 2022-06-24
extends logs with extra metadata and query params

r0wdy1:
This PR addresses the issue below
https://github.com/tomusdrw/rust-web3/issues/650

### DIFF
--- a/src/contract/mod.rs
+++ b/src/contract/mod.rs
@@ -319,9 +319,7 @@ impl<T: Transport> Contract<T> {
                     data: l.data.0,
                 })?;
 
-                Ok(R::from_tokens(
-                    log.params.into_iter().map(|x| x.value).collect::<Vec<_>>(),
-                )?)
+                R::from_tokens(log.params.into_iter().map(|x| x.value).collect::<Vec<_>>())
             })
             .collect::<Result<Vec<R>>>()
     }
--- a/src/contract/mod.rs
+++ b/src/contract/mod.rs
@@ -371,6 +371,15 @@ mod contract_signing {
                 // TODO [ToDr] SendTransactionWithConfirmation should support custom error type (so that we can return
                 // `contract::Error` instead of more generic `Error`.
                 .map_err(|err| crate::error::Error::Decoder(format!("{:?}", err)))?;
+            self.sign_raw(fn_data, options, key).await
+        }
+
+        async fn sign_raw(
+            &self,
+            fn_data: Vec<u8>,
+            options: Options,
+            key: impl signing::Key,
+        ) -> crate::Result<SignedTransaction> {
             let accounts = Accounts::new(self.eth.transport().clone());
             let mut tx = TransactionParameters {
                 nonce: options.nonce,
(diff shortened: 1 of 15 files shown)


src/contract/mod.rs
```diff
@@ -279,7 +279,16 @@ impl<T: Transport> Contract<T> {
     }
 
     /// Find events matching the topics.
```
tomusdrw, line 281: The documentation should be updated.

src/types/log.rs
```diff
@@ -54,6 +57,17 @@ impl Log {
     }
 }
 
+/// A log produced when a specefic contract event was emited
```
tomusdrw, line 60:
```suggestion
/// A log produced when a specific contract event was emitted.
```

src/types/log.rs
```diff
@@ -54,6 +57,17 @@ impl Log {
...
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
+pub struct LogWithMeta<R: Detokenize> {
+    /// This could be necessary to retrieve calldata
+    #[serde(rename = "transactionHash")]
+    pub transaction_hash: Option<H256>,
+
+    ///this is a tuple from the event signature (e.g. (uint256, bytes))
```
tomusdrw, line 67:
```suggestion
    /// A tuple from the event signature (e.g. `(uint256, bytes)`).
```

src/types/log.rs
```diff
@@ -54,6 +57,17 @@ impl Log {
     }
 }
 
+/// A log produced when a specefic contract event was emited
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
+pub struct LogWithMeta<R: Detokenize> {
+    /// This could be necessary to retrieve calldata
```
tomusdrw, line 63:
```suggestion
    /// Transaction hash necessary to retrieve `calldata`.
```

src/types/log.rs
```diff
@@ -106,27 +120,33 @@ impl FilterBuilder {
...
     /// mutually exclusive. Setting `from_block` will clear a previously set
     /// `block_hash`.
-    pub fn from_block(mut self, block: BlockNumber) -> Self {
-        self.filter.block_hash = None;
-        self.filter.from_block = Some(block);
+    pub fn from_block(mut self, block: Option<BlockNumber>) -> Self {
+        if let Some(block) = block {
+            self.filter.block_hash = None;
+            self.filter.from_block = Some(block);
+        }
```
tomusdrw, line 127: This method should not be changed, the builder should only take values that are defined, not `Option`s. `None` case should be handled at the call site.

src/types/log.rs
```diff
@@ -106,27 +120,33 @@ impl FilterBuilder {
...
     /// Sets `to_block`. The fields `to_block` and `block_hash` are mutually
     /// exclusive. Setting `to_block` will clear a previously set `block_hash`.
-    pub fn to_block(mut self, block: BlockNumber) -> Self {
-        self.filter.block_hash = None;
-        self.filter.to_block = Some(block);
+    pub fn to_block(mut self, block: Option<BlockNumber>) -> Self {
+        if let Some(block) = block {
+            self.filter.block_hash = None;
+            self.filter.to_block = Some(block);
+        }
```
tomusdrw, line 137: Same here.

src/types/log.rs
```diff
@@ -106,27 +120,33 @@ impl FilterBuilder {
...
-    pub fn block_hash(mut self, hash: H256) -> Self {
-        self.filter.from_block = None;
-        self.filter.to_block = None;
-        self.filter.block_hash = Some(hash);
+    pub fn block_hash(mut self, hash: Option<H256>) -> Self {
+        if let Some(_block_hash) = hash {
+            self.filter.from_block = None;
+            self.filter.to_block = None;
+            self.filter.block_hash = hash;
+        }
```
tomusdrw, line 149: ditto.

src/contract/mod.rs
```diff
@@ -279,7 +279,16 @@ impl<T: Transport> Contract<T> {
...
-    pub async fn events<A, B, C, R>(&self, event: &str, topic0: A, topic1: B, topic2: C) -> Result<Vec<R>>
+    pub async fn events<A, B, C, R>(
+        &self,
+        event: &str,
+        from_block: Option<BlockNumber>,
+        to_block: Option<BlockNumber>,
+        block_hash: Option<H256>,
```
tomusdrw, line 287: Having 7 parameters, and 3 of them being optional is a bit too much. I'd like to propose a helper struct with a builder to amalgamate them.
```rust
pub struct BlockRange {
  from_block: Option<BlockNumber>,
  to_block: Option<BlockNumber>,
  block_hash: Option<H256>,
}

impl BlockRange {
  // set both `from_block` and `to_block`
  pub fn range(from: BlockNumber, to: BlockNumber) -> Self;
  // set just `from_block`
  pub fn from(BlockNumber) -> Self;
  // set just `to_block`
  pub fn to(BlockNumber) -> Self;
  // set just `block_hash`
  pub fn hash(H256) -> Self;
  // look for one exact block (set both `from/to` and `hash`).
  pub fn exact(BlockNumber, H256) -> Self;

  // applies the parameters to given filter
  pub (crate) fn apply_to(&self, b: FilterBuilder) -> FilterBuilder;
}
```

tomusdrw (changes requested):

PR #656, opened 2022-07-06
Handling the case of empty return type

sergeyT2:


tomusdrw (changes requested):
Needs tests and some in-code documentation to explain what kind of use case it's solving.

PR #657, opened 2022-07-08
Add the wallet_switchEthereumChain method to the eip1193 transport

monomadic:
This adds support for [eip3326](https://eips.ethereum.org/EIPS/eip-3326), calling `wallet_switchEthereumChain`. It will allow a dapp to request a chain switch for a user with a eip1193 transport (such as metamask).

There are other methods that go alongside this method, but I'm not sure if this is the canonical way to implement this into `rust-web3`, so I only implemented the smallest (simplest) method first.

To test this, you'd want to have a transport from your provider like this:
```rust
        let transport = Eip1193::new(self.provider.clone());
        transport.switch_chain(chain_id)
```

I'm not sure if the return values are ideomatic to your api either. Any guidance would be helpful as these functions are critical for any user wishing to develop web3 dapps in rust with a eip1193 transport.

### DIFF
--- a/src/transports/eip_1193.rs
+++ b/src/transports/eip_1193.rs
@@ -132,6 +132,24 @@ impl Eip1193 {
             .into_deserializer();
         O::deserialize(deserializer).expect(&format!("couldn't deserialize {}", name))
     }
+
+    /// EIP-3326: Switch a wallet to another chain
+    pub async fn switch_chain(&self, chain_id: &str) -> Result<serde_json::value::Value, error::Error> {
+        let js_params = JsValue::from_serde(&vec![&ChainId { chain_id: chain_id.to_string() }])
+            .expect("couldn't send method params via JSON");
+
+        self.provider_and_listeners.borrow().provider.request_wrapped(RequestArguments {
+            method: String::from("wallet_switchEthereumChain"),
+            params: js_sys::Array::from(&js_params),
+        })
+        .await
+    }
+}
+
+#[derive(serde::Serialize)]
+#[serde(rename_all = "camelCase")]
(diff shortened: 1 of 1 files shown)


monomadic:
is this repository maintained any more? seems like there's open PRs from months back...

src/transports/eip_1193.rs
```diff
@@ -132,6 +132,29 @@ impl Eip1193 {
...
+                method: String::from("wallet_switchEthereumChain"),
+                params: js_sys::Array::from(&js_params),
+            })
+            .await
+    }
+}
+
+#[derive(serde::Serialize)]
+#[serde(rename_all = "camelCase")]
+pub struct ChainId {
```
tomusdrw, line 156: Please move the struct declaration into the function, since I don't think it's required anywhere outside it (or just use `json!` internally)

src/transports/eip_1193.rs
```diff
@@ -132,6 +132,29 @@ impl Eip1193 {
             .into_deserializer();
         O::deserialize(deserializer).expect(&format!("couldn't deserialize {}", name))
     }
+
+    /// EIP-3326: Switch a wallet to another chain
+    pub async fn switch_chain(&self, chain_id: &str) -> Result<serde_json::value::Value, error::Error> {
+        let js_params = JsValue::from_serde(&vec![&ChainId {
+            chain_id: chain_id.to_string(),
+        }])
```
tomusdrw, line 140: might be easier to use `json!` macro for such a simple parameter.

tomusdrw (changes requested):

PR #668, opened 2022-11-04
fix: re-export rpc::Call

semtexzv:
This type is used in the public API of the crate. Re-exporting seems like a good idea.


tomusdrw (changes requested):
I think I'd rather see it re-exported from the `types` module. Please also add `Value` type, for the exact same reason.

PR #687, opened 2023-04-21
Add support for rustls websocket connection as feature `ws-rustls-tokio`

alexheretic:
Add feature `ws-rustls-tokio` as an alternative to `ws-tls-tokio`. This allows avoiding openssl similar to the existing `http-rustls-tls` feature.

As a new additional feature this is a non-breaking change.

### DIFF
--- a/README.md
+++ b/README.md
@@ -117,8 +117,10 @@ web3 = { version = "_", default-features = false, features = ["http"] }
 On Linux, `native-tls` is implemented using OpenSSL. To avoid that dependency
 for HTTPS or WSS use the corresponding features.
 ```toml
-web3 = { version = "_", default-features = false, features = ["http-rustls-tls", "signing", "ws-rustls-tokio", "ipc-tokio"] }
+web3 = { version = "_", default-features = false, features = ["http-rustls-tls", "ws-rustls-tokio"] }
 ```
+
+_Note: To fully replicate the default features also add `signing` & `ipc-tokio` features_.
 
 # Cargo Features
 
--- a/README.md
+++ b/README.md
@@ -108,16 +108,16 @@ web3.api::<CustomNamespace>().custom_method().wait().unwrap()
 
 Currently, Windows does not support IPC, which is enabled in the library by default.
 To compile, you need to disable the IPC feature:
-```
-web3 = { version = "0.19.0", default-features = false, features = ["http"] }
+```toml
+web3 = { version = "_", default-features = false, features = ["http"] }
 ```
 
 # Avoiding OpenSSL dependency
 
 On Linux, `native-tls` is implemented using OpenSSL. To avoid that dependency
 for HTTPS use the corresponding feature.
-```
-web3 = { version = "0.19.0", default-features = false, features = ["http-rustls-tls"] }
+```toml
+web3 = { version = "_", default-features = false, features = ["http-rustls-tls", "signing", "ws-rustls-tokio", "ipc-tokio"] }
 ```
 
 # Cargo Features
@@ -129,6 +129,7 @@ The library supports following features:
 - `http-rustls-tls` - Enables TLS support via `reqwest/rustls-tls` for HTTP transport (implies `http`).
 - `ws-tokio` - Enables WS transport using `tokio` runtime.
 - `ws-tls-tokio` - Enables TLS support for WS transport (implies `ws-tokio`; default).
+- `ws-rustls-tokio` - Enables rustls TLS support for WS transport (implies `ws-tokio`).
 - `ws-async-std` - Enables WS transport using `async-std` runtime.
 - `ws-tls-async-std` - Enables TLS support for WS transport (implies `ws-async-std`).
 - `ipc-tokio` - Enables IPC transport using `tokio` runtime (default).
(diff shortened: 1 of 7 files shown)


alexheretic:
Hey @tomusdrw this would be useful for avoiding openssl, wdyt?

README.md
```diff
@@ -108,16 +108,16 @@ web3.api::<CustomNamespace>().custom_method().wait().unwrap()
...
-```
-web3 = { version = "0.17.0", default-features = false, features = ["http"] }
+```toml
+web3 = { version = "_", default-features = false, features = ["http"] }
 ```
 
 # Avoiding OpenSSL dependency
 
 On Linux, `native-tls` is implemented using OpenSSL. To avoid that dependency
 for HTTPS use the corresponding feature.
```
tomusdrw, line 118:
```suggestion
for HTTPS or WSS use the corresponding features.
```

README.md
```diff
@@ -108,16 +108,16 @@ web3.api::<CustomNamespace>().custom_method().wait().unwrap()
...
-```
-web3 = { version = "0.17.0", default-features = false, features = ["http-rustls-tls"] }
+```toml
+web3 = { version = "_", default-features = false, features = ["http-rustls-tls", "signing", "ws-rustls-tokio", "ipc-tokio"] }
```
tomusdrw, line 120: Why add `signing` and `ipc-tokio` here? The purpose of this example is to explain how to avoid OpenSSL.
```suggestion
web3 = { version = "_", default-features = false, features = ["http-rustls-tls", "ws-rustls-tokio"] }
```

src/transports/ws.rs
```diff
@@ -134,8 +134,17 @@ impl WsServerTask {
                 let stream = async_native_tls::connect(host, stream).await?;
                 MaybeTlsStream::Tls(compat::compat(stream))
             }
-            #[cfg(not(any(feature = "ws-tls-tokio", feature = "ws-tls-async-std")))]
-            panic!("The library was compiled without TLS support. Enable ws-tls-tokio or ws-tls-async-std feature.");
+            #[cfg(all(
+                feature = "ws-rustls-tokio",
+                not(feature = "ws-tls-tokio"),
+                not(feature = "ws-tls-async-std")
```
tomusdrw, line 140: I think it might be a bit confusing to fallback to `ws-tls-tokio` in case multiple features are enabled. I think I'd rather explicitly panic here or ideally during compilation.

src/transports/ws.rs
```diff
@@ -512,13 +548,16 @@ pub mod compat {
     /// TLS stream type for tokio runtime.
     #[cfg(feature = "ws-tls-tokio")]
     pub type TlsStream = Compat<async_native_tls::TlsStream<tokio::net::TcpStream>>;
+    /// Rustls TLS stream type for tokio runtime.
+    #[cfg(all(feature = "ws-rustls-tokio", not(feature = "ws-tls-tokio")))]
```
tomusdrw, line 552:
```suggestion
    #[cfg(feature = "ws-rustls-tokio")]
```

After we explicitly panic during build time when both `ws-tls-tokio` and `ws-rustls-tokio` are enabled.

tomusdrw (changes requested):
Looks cool, thanks! I left a couple of suggestions though.

README.md
alexheretic: This replicates the default features replacing openssl with rustls.

src/transports/ws.rs
alexheretic: That approach can work, but in less simple use cases it isn't always ideal. In deeper dependency trees you could have a dependency crate _A_ using _web3_ with default features, and then use it top-level with rustls.

Considering that case:
* A runtime failure seems undesirable and unexpected.
* A build error makes more sense. But can be painful to address. From the perspective of top-level you need to push _rustls_ changes upstream or fork. For the perspective of the crate _A_ maintainer changing openssl->rustls is more of a breaking change (since it can cause downstream builds to fail).
* Whereas here _web3_ continues working and just chooses _openssl_ code since both features were selected. It's easier to maintain from the outside since crate _A_ can change, e.g. to _rustls_ by default without risking breaking downstream. It is up to the user to actually ensure openssl isn't selected by any dependency, but things will generally still work.

We could also could prefer _rustls_, instead of openssl, when both are selected. I just picked what was already there for the preference.

Out of all of those Just let me know which way you'd prefer to go :slightly_smiling_face:

README.md
alexheretic: I've changed it to mentioning the default features in an additional note below.

src/transports/ws.rs
alexheretic: @tomusdrw I'm happy to handle this case with a compile error or working (preferring one or the other), can you let me know which you want & I'll update PR.

tomusdrw: @alexheretic thanks for explanations and sorry for the delay. I think it's fine the way it is, your explanation makes a lot of sense.

tomusdrw (approved):

alexheretic:
@tomusdrw I've updated the _tokio-rustls_ & _webpki-roots_ dependencies. Should be ready to go now.

alexheretic:
I've updated to the latest _rustls_. Merge?

tomusdrw (approved):

alexheretic:
Any chance we could get a new release with this feature?

PR #695, opened 2023-06-14
Fix Bytes deserialization for unicode values

popzxc:
Current `Bytes` deserialization logic can panic if supplied non-ASCII values: it uses [`str::len`](https://doc.rust-lang.org/std/primitive.str.html#method.len) and [`SliceIndex<str>`](https://doc.rust-lang.org/std/primitive.str.html#impl-SliceIndex%3Cstr%3E-for-Range%3Cusize%3E), both of which are byte-based (not char-based).
The deserializer code can panic when `[0..2]` range is not on the char boundary.

This PR fixed it by checking whether the string is ASCII first.

### DIFF
--- a/src/types/bytes.rs
+++ b/src/types/bytes.rs
@@ -54,8 +54,8 @@ impl<'a> Visitor<'a> for BytesVisitor {
     where
         E: Error,
     {
-        if value.len() >= 2 && &value[0..2] == "0x" {
-            let bytes = hex::decode(&value[2..]).map_err(|e| Error::custom(format!("Invalid hex: {}", e)))?;
+        if let Some(value) = value.strip_prefix("0x") {
+            let bytes = hex::decode(value).map_err(|e| Error::custom(format!("Invalid hex: {}", e)))?;
             Ok(Bytes(bytes))
         } else {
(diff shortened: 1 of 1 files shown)


src/types/bytes.rs
```diff
@@ -54,7 +54,7 @@ impl<'a> Visitor<'a> for BytesVisitor {
     where
         E: Error,
     {
-        if value.len() >= 2 && &value[0..2] == "0x" {
+        if value.is_ascii() && value.len() >= 2 && &value[0..2] == "0x" {
```
tomusdrw, line 57:
```suggestion
        if value.starts_with("0x") {
```

Wouldn't actually that work better?

tomusdrw (commented):
Good catch! I proposed a change to maybe just use `starts_with`, can you check if it passes all the tests?

src/types/bytes.rs
popzxc: Oh, indeed! However, it can be expressed even better with `str::strip_prefix`. Pushed the change.

tomusdrw (approved):

popzxc:
@tomusdrw I kind of need this fix for my project, could you please release a new version?

Armss9936 (approved):

PR #703, opened 2023-08-07
Add wasm support to transports::Either

platonfloria:
Issue:
'the trait `web3::Transport` is not implemented for `web3::transports::Either' when used in wasm.

Reason:
wasm Futures do not implement Send, that is required for Transport implementation of Either.
See https://github.com/rustwasm/wasm-bindgen/issues/2833

Solution:
Remove Send requirement for wasm implementation by using LocalBoxFuture.


tomusdrw (changes requested):
Can you elaborate why the transport does not work in `wasm` currently?
I have to admit I don't fully understand neither the problem nor the solution, so would appreciate some insights in the PR description.

platonfloria:
> Can you elaborate why the transport does not work in `wasm` currently? I have to admit I don't fully understand neither the problem nor the solution, so would appreciate some insights in the PR description.

Done

PR #711, opened 2023-11-21
test(types): add test should_deserialize_receipt_with_logs

pluveto:
Add a log deserialization test.
Added support for dynamic listening ports to avoid occasional failures during concurrent tests


tomusdrw (approved):

tomusdrw:
Thanks!

PR #717, opened 2024-02-05
Feature: add revert reason to receipt

DenisRybas:
Changes:
- Added optional `revert_reason` String to receipt

Reasoning: 
`revert(...)`, as described in [Solidity docs](https://docs.soliditylang.org/en/latest/control-structures.html#revert), can return a revert reason as an ABI-encoded string. Some Ethereum clients already have this functionality (e.g. [Hyperledger Besu](https://besu.hyperledger.org/private-networks/how-to/send-transactions/revert-reason)), while in others it may be added later.

### DIFF
--- a/src/confirm.rs
+++ b/src/confirm.rs
@@ -4,10 +4,16 @@ use crate::{
     api::{Eth, EthFilter, Namespace},
     error,
     types::{Bytes, TransactionReceipt, TransactionRequest, H256, U64},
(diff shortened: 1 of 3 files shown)


Toktar:
Hello @tomusdrw ,
could you please review and merge this PR?
Thanks in advance!

tomusdrw (approved):

PR #731, opened 2024-07-01
FilterBuilder::topics documentation 

mohamed-said:
## Overview

Add explanatory documentation to the FilterBuilder::topics method, as it takes up to 4 arguments. To make it easy for the users to know how to use it and what to expect.


tomusdrw (approved):
thanks!

Issue #397, opened 2020-10-27
Do not unwrap transaction receipt during confirmation


e00E:
https://github.com/tomusdrw/rust-web3/blob/8be891569c83e6f174bf7994ba36051c79858d32/src/confirm.rs#L277 should not `expect`. No matter what happens inside the node, this shouldn't be a panic.

Could convert this into some error. InternalError probably makes sense but then we cannot provide the error message. Maybe make new error for this? Or allow InternalError to carry a message.

YeautyYE:
When will it be repaired

Issue #396, opened 2020-10-27
Add additional checks to block filter during confirmation


e00E:
https://github.com/tomusdrw/rust-web3/blob/8be891569c83e6f174bf7994ba36051c79858d32/src/confirm.rs#L66 should have additional checks:
* We should not continue calling next on a stream that has completed (has returned None). We expect this to never happen for the blocks filter but to be safe we should handle this case for example by `fuse`ing the stream or
 erroring when it does complete.
* We do not handle the case where the stream returns an error which means we are wrongly counting it as a confirmation.

PR #740, opened 2025-10-24
Fix TODO #396 and #397: safe stream handling & receipt handling

aman879:
- Added safe handling for block streams in wait_for_confirmations
- Properly handle None receipt instead of panic in send_transaction_with_confirmation_
- Uses ok_or_else instead of expect
- Ensures no panics in production scenarios


tomusdrw (approved):
lgtm, thanks!

